PR #3, opened 2020-08-25
Fix COVID19 typo

todeveni:


travenin:
Similar typo seems to be in exposure-notification's pom.xml:
https://github.com/THLfi/koronavilkku-backend/blob/7d9cb0f7fae5fb5a9bc6d80567efc3597acac7f1/exposure-notification/pom.xml#L19

jyrkija (approved):

jyrkija:
Good catch. Will fix the other one as well.

PR #8, opened 2020-08-25
SQL readability improvements

mtarvainen:
I think title says it all. It is more easy to read when SQL is written in capitals :)


jyrkija:
It's indeed a very common practice to use caps for keywords, but as IDEs these days do syntax highlighting for SQL, it has become a subjective matter of preference. With a quick internal poll, we got about equal preference for either style and hence our current style guide for the project is that SQL keywords are not written in caps.

More importantly, these DAO-classes are not the only ones with SQL. The same practice is followed systematically in the migration SQLs (src/main/resources/db/... in both projects) and changing those would be breaking for current installations and a bit risky this close to release. We think it's more important to stick to a single consistent style than to switch for any particular one.

Thank you for your interest in the project though, and I'm sorry for the unnecessary work. We'll get the style guide up and visible shortly to avoid such things in the future.

mtarvainen:
I agree to disagree and I understand your worries about the close release date. Thanks for you reply though and all the best for upcoming release.

PR #19, opened 2020-09-02
Implementation of debug logging for validate only requests

solita-alperttiti:


publish-token/src/main/java/fi/thl/covid19/publishtoken/Validation.java
```diff
@@ -16,34 +16,41 @@ private Validation() {}
 
     public static String validatePublishToken(String publishToken) {
         if (!publishToken.matches(TOKEN_REGEX)) {
-            throw new InputValidationException("Invalid publish token.");
+            throw new InputValidationException("Invalid publish token.", false);
```
jyrkija: Pushing this information into the exceptions and validations seems overly intrusive. 
If we would have this as a header, then we could just pick it up in the ApiErrorHandler, from the WebRequest arg in all functions. I wonder if we'd have some other, cleaner alternatives as well...
I suppose we could push it into the MDC context also, though I don't think that's really pretty either (just a global variable really).

jyrkija: Actually, MDC won't even work since we get the attribute only after parsing is done.

publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationRequest.java
```diff
@@ -23,12 +24,22 @@ public PublishTokenGenerationRequest(String requestUser,
...
+        try {
+            this.requestUser = validateUserName(requireNonNull(requestUser, "User required"));
+            this.symptomsOnset = requireNonNull(symptomsOnset, "Symptoms onset date required");
+            this.patientSmsNumber = requireNonNull(patientSmsNumber).map(Validation::normalizeAndValidatePhoneNumber);
+            if (symptomsOnset.isAfter(LocalDate.now(ZoneOffset.UTC).plus(1, ChronoUnit.DAYS))) {
+                throw new InputValidationException("Symptoms onset time in the future: date=" + symptomsOnset);
+            }
+        } catch (InputValidationException | NullPointerException e) {
+            if (this.validateOnly) {
+                throw new InputValidationValidateOnlyException(e.getMessage());
```
jyrkija: The original exception could be put in as the cause -throwable, so that the stack (if ever used) is correct and points to the row where the problem happened. In this case, it doesn't matter of course, because we're never going to log that stack.

jyrkija (approved):

jyrkija (approved):

PR #20, opened 2020-09-02
Sanitizing InputValidationValidateOnlyException handling

solita-alperttiti:


publish-token/src/main/java/fi/thl/covid19/publishtoken/error/ApiErrorHandler.java
```diff
@@ -2,6 +2,7 @@
 
 import io.micrometer.core.lang.Nullable;
 import org.apache.catalina.connector.ClientAbortException;
+import org.flywaydb.core.internal.util.ExceptionUtils;
```
jyrkija: This dependency is some internal thing from under flyway. Better find a more reliable one to use and preferable one that doesn't tie into flyway.

jyrkija (changes requested):

publish-token/src/main/java/fi/thl/covid19/publishtoken/error/ApiErrorHandler.java
solita-alperttiti: Yep, this was already simultaneously fixed. :)

jyrkija (approved):

PR #23, opened 2020-09-07
Preliminary implementation for usage statistics

solita-alperttiti:
This pull request resolves COVID-380 .

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
@@ -57,6 +57,7 @@ public class DiagnosisKeyDao {
     public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys) {
         if (verify(verificationId, requestChecksum) && !keys.isEmpty()) {
             batchInsert(interval, keys);
+            addDiagnosisKeyStatsRow(Instant.now());
             LOG.info("Inserted keys: {} {}", keyValue("interval", interval), keyValue("count", keys.size()));
         }
     }
@@ -131,6 +132,13 @@ public class DiagnosisKeyDao {
                 .map(key -> createParamsMap(interval, key))
                 .toArray((IntFunction<Map<String, Object>[]>) Map[]::new);
         jdbcTemplate.batchUpdate(sql, params);
+    }
+
+    private void addDiagnosisKeyStatsRow(Instant createTime) {
+        String sql = "insert into en.stats_diagnosis_keys_created(created_at) values (:created_at)";
+        Map<String, Object> params = Map.of(
+                "created_at", new Timestamp(createTime.toEpochMilli()));
+        jdbcTemplate.update(sql, params);
     }
 
     private Optional<TemporaryExposureKey> mapValidKey(int interval, ResultSet rs, int index) throws SQLException {
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
@@ -57,7 +57,7 @@ public class DiagnosisKeyDao {
     public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys) {
         if (verify(verificationId, requestChecksum) && !keys.isEmpty()) {
             batchInsert(interval, keys);
-            addDiagnosisKeyStatsRow(Instant.now());
+            addReportKeysStatsRow(Instant.now());
             LOG.info("Inserted keys: {} {}", keyValue("interval", interval), keyValue("count", keys.size()));
         }
     }
@@ -134,10 +134,10 @@ public class DiagnosisKeyDao {
         jdbcTemplate.batchUpdate(sql, params);
     }
 
-    private void addDiagnosisKeyStatsRow(Instant createTime) {
-        String sql = "insert into en.stats_diagnosis_keys_created(created_at) values (:created_at)";
+    private void addReportKeysStatsRow(Instant createTime) {
+        String sql = "insert into en.stats_report_keys(reported_at) values (:reported_at)";
         Map<String, Object> params = Map.of(
-                "created_at", new Timestamp(createTime.toEpochMilli()));
+                "reported_at", new Timestamp(createTime.toEpochMilli()));
         jdbcTemplate.update(sql, params);
     }
 
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
@@ -57,7 +57,6 @@ public class DiagnosisKeyDao {
     public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys) {
         if (verify(verificationId, requestChecksum) && !keys.isEmpty()) {
             batchInsert(interval, keys);
-            addReportKeysStatsRow(Instant.now());
             LOG.info("Inserted keys: {} {}", keyValue("interval", interval), keyValue("count", keys.size()));
         }
     }
@@ -110,6 +109,7 @@ public class DiagnosisKeyDao {
         boolean rowCreated = jdbcTemplate.update(sql, new MapSqlParameterSource(params)) == 1;
         LOG.info("Marked token verification: {}", keyValue("newVerification", rowCreated));
         if (rowCreated) {
+            addReportKeysStatsRow(Instant.now());
             return true;
         } else if (requestChecksum.equals(getVerifiedChecksum(verificationId))) {
             return false;
--- a/exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
+++ b/exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
@@ -1,4 +1,4 @@
-create table en.stats_diagnosis_keys_created (
+create table en.stats_report_keys (
     id bigint primary key generated always as identity,
-    created_at timestamptz not null
+    reported_at timestamptz not null
 );
--- a/exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
+++ b/exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
@@ -1,0 +1,4 @@
+create table en.stats_diagnosis_keys_created (
+    id bigint primary key generated always as identity,
+    created_at timestamptz not null
+);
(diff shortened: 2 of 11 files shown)


publish-token/src/main/java/fi/thl/covid19/publishtoken/sms/SmsService.java
```diff
@@ -30,18 +32,22 @@
...
                 keyValue("active", config.gateway.isPresent()),
                 keyValue("senderName", config.senderName),
                 keyValue("gateway", config.gateway.orElse("")));
     }
 
     public boolean send(String number, PublishToken token) {
-        if (config.gateway.isPresent()) {
-            return send(config.gateway.get(), number, config.formatContent(token.token));
+        if (config.gateway.isPresent() && send(config.gateway.get(), number, config.formatContent(token.token))) {
+            dao.addStatsRow(Instant.now(), "stats_sms_created");
```
jyrkija: I think only the DAO should know about table names. I guess you could make an enum argument for the "type of stats row", or alternatively, just add 2 methods for adding the stat.

publish-token/src/test/java/fi/thl/covid19/publishtoken/SMSServiceTest.java
```diff
@@ -41,6 +41,9 @@
     @MockBean
     private NamedParameterJdbcTemplate jdbc;
 
+    @MockBean
+    private PublishTokenDao publishTokenDao;
```
jyrkija: Rather than mock the dao, I think it would be better to turn this into an -IT test, mock nothing and verify that the stats-row shows up in DB as well.

publish-token/src/main/java/fi/thl/covid19/publishtoken/PublishTokenDao.java
```diff
@@ -78,6 +80,13 @@ public void deleteTokensExpiredBefore(Instant expiryLimit) {
         LOG.info("Publish tokens deleted: {} {}", keyValue("expiryLimit", expiryLimit.toString()), keyValue("count", count));
     }
 
+    public void addStatsRow(Instant createTime, String tableName) {
+        String sql = "insert into pt." + tableName + "(created_at) values (:created_at)";
```
jyrkija: Concatenating args into the sql is something of a bad habit. It doesn't create any issues here of course, but you do lose content assist on the table names. 

I think I'd personally just throw separate functions, eg. addSmsStat(Instant) and addTokenGenerationStat(Instant), especially since they're called from outside the class (now someone other than Dao needs to get the tablename right, without content assist). Of course that means that there's a touch of duplication, so it's a bit of a compromise.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDaoIT.java
```diff
@@ -46,12 +52,15 @@ public void createReadDeleteWorks() {
 
         dao.addKeys(1, md5DigestAsHex("test".getBytes()), interval, keys);
         assertKeysStored(interval, keys);
+        assertStatRowAdded();
 
         dao.deleteKeysBefore(interval);
         assertKeysStored(interval, keys);
+        assertStatRowAdded();
```
jyrkija, line 59: Does this work? Seems that there would be 2 stat-rows at this point and the assert says 1

publish-token/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
```diff
@@ -0,0 +1,9 @@
+create table pt.stats_tokens_created (
+    id bigint primary key generated always as identity,
+    created_at timestamptz not null
+);
+
+create table pt.stats_sms_created (
```
jyrkija: Since we have separate tables, sms_sent (table) and sent_at (column) or send_time would sound more logical.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -133,6 +134,13 @@ private void batchInsert(int interval, List<TemporaryExposureKey> newKeys) {
         jdbcTemplate.batchUpdate(sql, params);
     }
 
+    private void addDiagnosisKeyStatsRow(Instant createTime) {
+        String sql = "insert into en.stats_diagnosis_keys_created(created_at) values (:created_at)";
+        Map<String, Object> params = Map.of(
```
jyrkija, line 139: In these trivial param-maps, I've usually just given it directly in ...update(sql, Map.of("created_at", ...)
This is fine too, of course

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -57,6 +57,7 @@ public int deleteVerificationsBefore(Instant verificationTime) {
     public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys) {
         if (verify(verificationId, requestChecksum) && !keys.isEmpty()) {
             batchInsert(interval, keys);
+            addDiagnosisKeyStatsRow(Instant.now());
```
jyrkija: This won't be called when someone makes a valid push-request with a valid token but 0 keys (fresh installation). I think those should be included in the stat, as it will be compared to publish token creation stats. Hence, it might be better to call it from within verify(..) method. As such, it could also be named "stats_token_verified" or somesuch.

exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
```diff
@@ -0,0 +1,4 @@
+create table en.stats_diagnosis_keys_created (
```
jyrkija: This is a slightly misleading name, as it logs post-events and each event may create 0-14 keys.

jyrkija (changes requested):

publish-token/src/main/java/fi/thl/covid19/publishtoken/sms/SmsService.java
solita-alperttiti: I don't like repeating myself, and I have happily lived with project with dao and service combined. But, if you insist I can make enum in dao. :)

publish-token/src/test/java/fi/thl/covid19/publishtoken/SMSServiceTest.java
solita-alperttiti: Yep, that might be good, but decided that testing writing in dao tests is enough.

publish-token/src/main/java/fi/thl/covid19/publishtoken/PublishTokenDao.java
solita-alperttiti: I don't like the idea. It's not bad habit if you don't pass some userinput. However, your point about assist is valid, but I still prefer not to duplicate code, even if it's only few lines, like in this case.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDaoIT.java
solita-alperttiti: There is only one store operation, so only one stats row. Have I misunderstood something?

publish-token/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
solita-alperttiti: Yep, good point. I'll change that.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: That is matter of taste, obviously. :)

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Ok, good catch. I'll fix that

exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
solita-alperttiti: Yep, it refers to api, not actual keys, but I'll make better name.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
jyrkija: Yup

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDaoIT.java
jyrkija: No, you're correct. I misread the test, it's fine as it is.

publish-token/src/main/java/fi/thl/covid19/publishtoken/sms/SmsService.java
jyrkija: Not necessary, it's not really that much better and it's a code-style nitpick anyhow. I dislike this, but I don't really have a better option. Like mentioned below, I think I'd prefer separate methods for separate stats, but that's a compromise.

solita-alperttiti: I have pushed separate methods. I was have it in that way, but it was ugly. But I'm not a fan to fight about style, so there it is.

publish-token/src/test/java/fi/thl/covid19/publishtoken/SMSServiceTest.java
solita-alperttiti: Fixed

publish-token/src/main/java/fi/thl/covid19/publishtoken/PublishTokenDao.java
solita-alperttiti: Changed it anyway. While generalization migh be good in general, in this case two separate method will also do. Until we have 100 stats tabels or something.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Fixed

exposure-notification/src/main/resources/db/migration/V02.01__create_statistics_tables.sql
solita-alperttiti: Changed

jyrkija (approved):

PR #24, opened 2020-09-09
Release automatization

solita-alperttiti:
This pull request resolves COVID-380.

### DIFF
--- a/create_release_branch.sh
+++ b/create_release_branch.sh
@@ -55,7 +55,6 @@ find . -name "pom.xml" -exec sed -i "s|<version>$CURRENT_VERSION</version>|<vers
 
 git commit -a -m "Rolled version for new development version"
 
-echo "Do you want to push new version to remote?"
 while true; do
     read -p "Do you want to push new version to remote?" yn
     case $yn in
--- a/create_release_branch.sh
+++ b/create_release_branch.sh
@@ -41,7 +41,7 @@ CURRENT_VERSION=$(grep -oPm1 "(?<=<version>)[^<]+" pom.xml)
 git checkout -b release/$CURRENT_VERSION
 
 while true; do
-    read -rp "Do you want to push new release branch to remote?" yn
+    read -rp "Do you want to push new release branch(release/$CURRENT_VERSION) to remote?" yn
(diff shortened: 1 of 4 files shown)


publish-token/pom.xml
```diff
@@ -4,17 +4,17 @@
 
 	<modelVersion>4.0.0</modelVersion>
 
+    <groupId>fi.thl.covid19</groupId>
```
jyrkija, line 7: different indentation?

exposure-notification/pom.xml
```diff
@@ -4,17 +4,17 @@
 
 	<modelVersion>4.0.0</modelVersion>
 
+    <groupId>fi.thl.covid19</groupId>
```
jyrkija, line 7: Indentation here too. Is this a github-thing or real difference?

create_release_branch.sh
```diff
@@ -0,0 +1,66 @@
...
+git fetch --all --tags --prune
+git checkout trunk
+git pull --rebase
+
+CURRENT_VERSION=$(grep -oPm1 "(?<=<version>)[^<]+" pom.xml)
+
+git checkout -b release/$CURRENT_VERSION
+
+while true; do
+    read -rp "Do you want to push new release branch to remote?" yn
```
jyrkija: Might be helpful to include the version(s) in this query text. Same thing below.

jyrkija (changes requested):

publish-token/pom.xml
solita-alperttiti: Good question. It seems to be fine in my end.

exposure-notification/pom.xml
solita-alperttiti: Good question. It seems to be fine in my end.

solita-alperttiti: No, there is tabs... I'll fix that

create_release_branch.sh
solita-alperttiti: Yep, this is a good one. Added.

jyrkija (approved):

PR #26, opened 2020-09-23
Allow validateOnly to be provided in header on V1 publishToken api

solita-alperttiti:
Opened this for discussion.

Resolves: COVID-403

### DIFF
--- a/publish-token/src/main/java/fi/thl/covid19/publishtoken/Validation.java
+++ b/publish-token/src/main/java/fi/thl/covid19/publishtoken/Validation.java
@@ -1,6 +1,12 @@
 package fi.thl.covid19.publishtoken;
 
+import com.fasterxml.jackson.databind.InjectableValues;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import fi.thl.covid19.publishtoken.error.InputValidationException;
+import fi.thl.covid19.publishtoken.error.InputValidationValidateOnlyException;
+import fi.thl.covid19.publishtoken.generation.v1.PublishTokenGenerationRequest;
+
+import java.io.IOException;
 
 public final class Validation {
     private Validation() {
@@ -14,6 +20,20 @@ public final class Validation {
     public static final int USER_NAME_MAX_LENGTH = 50;
     public static final int SERVICE_NAME_MAX_LENGTH = 100;
     private static final String NAME_REGEX = "([A-Za-z0-9\\-_.]+)";
+
+
+    public static PublishTokenGenerationRequest validateAndCreate(String jsonBody, boolean validateOnly, ObjectMapper om) {
+        InjectableValues injectableValues = new InjectableValues.Std().addValue(boolean.class, validateOnly);
+        try {
+            return om.reader(injectableValues).readValue(jsonBody, PublishTokenGenerationRequest.class);
+        } catch (IOException e) {
+            if (validateOnly) {
+                throw new InputValidationValidateOnlyException("ValidateOnly: " + e.getMessage());
+            } else {
+                throw new InputValidationException(e.getMessage());
+            }
+        }
+    }
 
     public static String validatePublishToken(String publishToken) {
         if (!publishToken.matches(TOKEN_REGEX)) {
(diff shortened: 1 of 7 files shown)


publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationRequest.java
```diff
@@ -23,9 +24,10 @@
     public PublishTokenGenerationRequest(String requestUser,
                                          LocalDate symptomsOnset,
                                          Optional<String> patientSmsNumber,
-                                         Optional<Boolean> validateOnly) {
+                                         Optional<Boolean> validateOnly,
+                                         @JacksonInject boolean validateOnlyHeader) {
```
jyrkija, line 28: Overall, this seems like a needlessly big/complex change just to add an optional header. Why not just pick the header in the controller and do something like request.validateOnly.orElseGet(validateOnlyHeader...)? 
Compared to this, the only difference I can tell is that the header never ends up in this object - and I don't see why it should. In fact I think it's cleaner if in the v2 version the whole validateOnly-logic is removed from the request-object and only handled by the controller.

publish-token/src/main/java/fi/thl/covid19/publishtoken/Validation.java
```diff
@@ -1,6 +1,7 @@
 package fi.thl.covid19.publishtoken;
 
 import fi.thl.covid19.publishtoken.error.InputValidationException;
+import org.springframework.web.context.request.NativeWebRequest;
```
jyrkija, line 4: I'd prefer to keep this class "clean" in the sense that it doesn't understand data sources, just data. Or in this case, not requests but just Strings.

publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationRequestArgumentResolver.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.web.context.request.NativeWebRequest;
+import org.springframework.web.method.support.HandlerMethodArgumentResolver;
+import org.springframework.web.method.support.ModelAndViewContainer;
+
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class PublishTokenGenerationRequestArgumentResolver implements HandlerMethodArgumentResolver {
+
+    public static final String VALIDATE_ONLY_HEADER = "X-Validate-Only";
```
jyrkija, line 20: We've used the prefix KV- for our other headers. The practice of using X- for custom headers is no-longer "the recommended standard" in general, though of course not forbidden either. I do agree that it's good to separate our own koronavilkku stuff with some prefix, which is how we ended up with KV-.

jyrkija (changes requested):

publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationRequest.java
solita-alperttiti: Design decision behind this approach is to combine header and body validate-only _before_ we'll do any de-serializing and such, because validateOnly is needed when constructing PublishTokenGenerationRequest to decide which kind of exception will be thrown. Therefore we cannot just do this in the controller. However, it's possible to do this in maybe a little bit less code by just checking header in controller  and do controller things based on that. Anyway, this means that then we'll need separate login in ErrorHandler to check header value as separated thing from current logic. From maintenance point of view this is not a very good thing, because coder needs to know that both places should be changed if there is some need for that. Anyway, you could do an poc (like we discussed) about that other approach and then we could decide which one is better. :)

publish-token/src/main/java/fi/thl/covid19/publishtoken/Validation.java
solita-alperttiti: Ok, I see it more like a util collection of static methods which do different kinds of validation related things. Anyway, this method can be moved to somehere else, there is no much sense for it if it just have data anyway, but let's see how this will be developing.

publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationRequestArgumentResolver.java
solita-alperttiti: Yep, good enough. I just wanted to separate actual client headers from others in some manner, but maybe it's really doesn't matter.

solita-alperttiti:
I'm closing this one. Let's go with the alternative version. Less code at least.

PR #27, opened 2020-09-24
Allow validateOnly to be provided in header on V1 publishToken API, alternative

jyrkija:
Alternative way of providing the backwards compatible validateonly header, not injecting it into the request object.

Resolves: COVID-403

### DIFF
--- a/publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationController.java
+++ b/publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationController.java
@@ -1,11 +1,13 @@
 package fi.thl.covid19.publishtoken.generation.v1;
 
 import fi.thl.covid19.publishtoken.PublishTokenService;
-import fi.thl.covid19.publishtoken.sms.SmsService;
 import fi.thl.covid19.publishtoken.Validation;
+import fi.thl.covid19.publishtoken.sms.SmsService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.web.bind.annotation.*;
+
+import java.util.Optional;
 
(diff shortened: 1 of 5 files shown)


publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationController.java
```diff
@@ -28,10 +31,12 @@ public PublishTokenGenerationController(PublishTokenService publishTokenService,
     }
 
     @PostMapping
-    public PublishToken generateToken(@RequestHeader(name = SERVICE_NAME_HEADER) String rawRequestService, @RequestBody PublishTokenGenerationRequest request) {
+    public PublishToken generateToken(@RequestHeader(name = SERVICE_NAME_HEADER) String rawRequestService,
+                                      @RequestHeader(name = VALIDATE_ONLY_HEADER) Optional<Boolean> validateOnly,
```
solita-alperttiti: This could be something like:

`@RequestHeader(name = VALIDATE_ONLY_HEADER, required = false) boolean validateOnly`

solita-alperttiti (changes requested):
In general, there is a less code, but dependency with controller and errorHandler is not visible in logic flow. Dunno, maybe this still might be a better approach, even though I don't like this at all. :) After corrections I vote for this one.

publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationController.java
jyrkija: Yeah, that's cleaner. There's a slight functional variation (if both are defined), but nothing relevant. I changed it to this.

solita-alperttiti (approved):

PR #28, opened 2020-09-28
Validate only documentation changes

solita-alperttiti:
This pull request resolves COVID-403.

### DIFF
--- a/publish-token/README.md
+++ b/publish-token/README.md
@@ -49,6 +49,8 @@ If the client provides a user-visible error and errorId from response, it can be
 ## Publish Token Generation API
 API meant for health authorities, for creating a publish-token for a diagnosis report after verifying the infection.
 
+Deprecated components will be removed in next api-version.
+
 ### Create a New Token after a Verified Diagnosis of COVID-19
 * **URL:** `/publish-token/v1`
 * **Method:** `POST`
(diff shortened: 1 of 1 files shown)


solita-alperttiti:
Should we start using CHANGELOG like we are recommending anyway? :)

publish-token/README.md
```diff
@@ -49,18 +49,21 @@ If the client provides a user-visible error and errorId from response, it can be
...
+Deprecated components will be removed in next api-version.
+
 ### Create a New Token after a Verified Diagnosis of COVID-19
 * **URL:** `/publish-token/v1`
 * **Method:** `POST`
 * **URL Params:** None
 * **Query Params:** None
 * **Headers:**
   * (Mandatory) `KV-Request-Service` Calling service name. (With default configuration this is resolved automatically on load balancer and calling client does not need to worry about this.)
+  * (Optional) `KV-Validate-Only` Boolean (true or false, if missing true). This is for API verification tests: request is validated and a token created, it isn't stored in the database (activated) or sent via SMS. (Overrides validateOnly in the body if present)
```
jyrkija: Doesn't strictly override with the current form - setting either to true will work, regardless the other.

jyrkija (approved):

publish-token/README.md
solita-alperttiti: Yep, meaning in here is actually that header is stronger than validateOnly if both are present. Also, if header is missing, then validate only applies if present.

jyrkija: That's what I mean. It doesn't work like that any more - the first version did, but when I changed it to be more spring-ish with an optional header bound to boolean (non-nullable) type, the condition is (header || field). So if the field is true, setting the header to false does not override. In practice, this makes no real difference, it's just semantics.

solita-alperttiti: Yes, you are correct.  Didn't think this to the end. Fixes pushed.

jyrkija (approved):

PR #31, opened 2020-10-07
Initial implementation for efgs integration

solita-alperttiti:
This pull request resolves partially COVID-412 and COVID-413 .


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -144,34 +126,76 @@ public boolean verify(int verificationId, String requestChecksum) {
...
-        String lock = "lock table en.efgs_update_to_operation in access exclusive mode";
+    public long startOutboundOperation() {
+        String lock = "lock table en.efgs_operation in access exclusive mode";
         jdbcTemplate.update(lock, Map.of());
-        if (isUpdateToEfgsAvailable()) {
-            return markUpdateToEfgsStarted();
+        if (isOutboundOperationAvailable()) {
+            return markOutboundOperationStarted();
         } else {
             throw new IllegalStateException("Update to efgs is already running.");
```
jyrkija: Probably want a special exception type here so that you can log it quietly in the error handler, as this is a business-as-usual exception.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -144,34 +126,76 @@ public boolean verify(int verificationId, String requestChecksum) {
...
+    public List<TemporaryExposureKey> fetchAvailableKeysForEfgs(long operationId) {
+        LOG.info("Fetching keys not sent to efgs.");
+        String sql = "select key_data, rolling_period, rolling_start_interval_number, transmission_risk_level, submission_interval " +
+                "from en.diagnosis_key " +
+                "where efgs_operation = :efgs_operation " +
+                "order by key_data";
+        return jdbcTemplate.queryForList(sql, Map.of("efgs_operation", operationId), TemporaryExposureKey.class);
+    }
+
+    public void finishOperation(long operationId) {
```
jyrkija: Maybe parametrize the new_state and combine with the below function

exposure-notification/pom.xml
```diff
@@ -100,7 +100,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
-
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
```
jyrkija, line 105: Is this needed? EN already does http-requests via resttemplate to PT, so I would have thought this is already included as transitive. If it is needed, the version should be attached as well.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+    private final Duration socketTimeout = Duration.ofSeconds(60);
+
+    private final String trustStorePath;
+    private final char[] trustStorePassword;
+    private final String keyStorePath;
+    private final char[] keyStorePassword;
+    private final String keyStoreKeyAlias;
+
+    public FederationGatewayRestClientConfiguration(
+            @Value("${covid19.federation-gateway.trust-store.path}") String trustStorePath,
```
jyrkija, line 37: You can set the long part of this property-key into the @Configuration as path-prefix to avoid repeating it

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -118,6 +125,81 @@ public boolean verify(int verificationId, String requestChecksum) {
...
+        jdbcTemplate.update(sql, Map.of("new_state", EfgsOperationState.FINISHED, "id", operationId));
+    }
+
+    public void markErrorOperation(long operationId) {
+        String sql = "update en.efgs_operation set state = :error_state where id = :id";
+        jdbcTemplate.update(sql, Map.of("error_state", EfgsOperationState.ERROR, "id", operationId));
+    }
+
+    @Transactional
+    public void addInboundKeys(List<TemporaryExposureKey> keys, int interval) {
```
jyrkija, line 134: If I understood this correctly, this doesn't seem like an optimal way to handle the locking for inbound data. We're fetching the data twice, then just failing the write, right? Inserting keys from 2 sources should be fine anyhow ("on conflict, do nothing") and we'd like to do the fetch just once.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -118,6 +125,81 @@ public boolean verify(int verificationId, String requestChecksum) {
...
+        if (!keys.isEmpty()) {
+            long operationId = startInboundOperation();
+            try {
+                batchInsert(interval, keys, operationId);
+                LOG.info("Inserted keys: {} {}", keyValue("interval", interval), keyValue("count", keys.size()));
+                finishOperation(operationId);
+            } catch (Throwable t) {
```
jyrkija: Never catch Throwable. Throwable includes JVM errors like out-of-memory etc. and there's no guarantee that your program is in any way runnable after it. Indeed, it might be something that's telling you to shut down on an OS level.
Prefer catching the most specific exception that makes sense and if you really need a wide catch, then make it Exception.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -118,6 +125,81 @@ public boolean verify(int verificationId, String requestChecksum) {
...
+                .stream().findFirst().orElseThrow(() -> new IllegalStateException("Count returned nothing.")) == 0;
+    }
+
+    private long markOutboundOperationStarted() {
+        KeyHolder operationKeyHolder = new GeneratedKeyHolder();
+
+        String createOperation = "insert into en.efgs_operation (direction) values (:direction)";
+        jdbcTemplate.update(createOperation, new MapSqlParameterSource("direction", EfgsOperationDirection.OUTBOUND), operationKeyHolder);
+
+        String markToBeUpdated = "update en.diagnosis_key " +
```
jyrkija: Personally, I'd rather try to avoid writing integration state into each key row. That's our major data mass and updating every row for an integration seems needlessly heavy... though it's probably not an actual performance issue due to small data amounts. This model also means that if we would have other integrations or stateful processors for the data, those would need their own state-columns and updates.

An optional way of doing this would be to include an insert-time in the key (maybe just replace the submission interval with a more accurate one - we'll need it later anyhow) and the integration status table would hold a field like processed_until. Of course you'd need to avoid processing too close to current time to avoid missing rows due to clock differences and ongoing transactions, but the keys themselves wouldn't need to know about integration states.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -128,12 +210,23 @@ private void batchInsert(int interval, List<TemporaryExposureKey> newKeys) {
...
                 "values (:key_data, :rolling_period, :rolling_start_interval_number, :transmission_risk_level, :submission_interval) " +
                 "on conflict do nothing";
-        Map<String,Object>[] params = newKeys.stream()
+        Map<String, Object>[] params = newKeys.stream()
                 .map(key -> createParamsMap(interval, key))
                 .toArray((IntFunction<Map<String, Object>[]>) Map[]::new);
         jdbcTemplate.batchUpdate(sql, params);
     }
 
+    private void batchInsert(int interval, List<TemporaryExposureKey> newKeys, long operationId) {
```
jyrkija, line 155: The only difference with above is the operationId, right? Maybe just make it an Optional<Long> parameter and combine the functions setting the parameter as operationId.orElse(null). Same with the two param-map-creation functions.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/error/EfgsOperationException.java
```diff
@@ -0,0 +1,7 @@
+package fi.thl.covid19.exposurenotification.error;
+
+public class EfgsOperationException extends RuntimeException {
+    public EfgsOperationException(String message, Throwable t) {
```
jyrkija: Maybe accept Exception instead of Throwable here too, just to make sure a potential mistake won't compile.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayClient.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public FederationGatewayClient(
+            @Qualifier("federationGatewayRestTemplate") RestTemplate restTemplate,
+            @Value("${covid19.federation-gateway.url-template}") String gatewayUrl
+    ) {
+        this.restTemplate = restTemplate;
+        this.gatewayUrl = gatewayUrl;
+    }
+
+    public int upload(String batchTag, String batchSignature, byte[] batchData) {
```
jyrkija: I would have guessed that interpreting status-codes would be FederationGatewayClient's responsibility, meaning that those could be processed here and the return value could be just OK/NOK (boolean). If it's the service's responsibility (which might make sense) then maybe it would be nicer to give out http statuscode object instead of an int.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayService.java
```diff
@@ -0,0 +1,114 @@
...
+        this.dd = diagnosisKeyDao;
+    }
+
+    public void doOutbound() {
+        long operationId = dd.startOutboundOperation();
+        try {
+            List<TemporaryExposureKey> localKeys = dd.fetchAvailableKeysForEfgs(operationId);
+            handleOutbound(transform(localKeys), operationId);
+            dd.finishOperation(operationId);
+        } catch (Throwable t) {
```
jyrkija: Don't catch throwable. Use finally-block for "whatever happens" operation closing. 

For instance, you could use a variable outside the block to mark success and then set operation end-result in a finally block with something like if(operationFinished) dd.finish... etc. Or maybe a parametrized function rather than 2 separate ones. Finally block also avoids the need to catch and rethrow anything as the original exception keeps on going.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayService.java
```diff
@@ -0,0 +1,114 @@
...
+            dd.markErrorOperation(operationId);
+            throw new EfgsOperationException("Outbound operation to efgs failed.", t);
+        }
+    }
+
+    public void doInbound(Optional<String> batchTag) {
+        // TODO: what date we should use?
+        String date = getDateString(LocalDate.now());
```
jyrkija: The result of this is dependent on the locale of the server. You can use LocalDate.now(ZoneOffset.UTC) to get something more predictable (the current UTC-date).

jyrkija (changes requested):

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -118,6 +125,81 @@ public boolean verify(int verificationId, String requestChecksum) {
         }
     }
 
+    @Transactional(timeout = 10)
+    public long startOutboundOperation() {
+        String lock = "lock table en.efgs_operation in access exclusive mode";
+        jdbcTemplate.update(lock, Map.of());
+        if (isOutboundOperationAvailable()) {
```
jyrkija: The operation reservations need a time-out as well. The service might shut-down in the middle of an update and there's no guarantee that the ending gets written

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
solita-alperttiti: Actually it would be good to have properties class for properties, but there was some discussion about that is  not good somwhow, so now it's in that way.

exposure-notification/pom.xml
solita-alperttiti: There was some reason, in some phase, but might not be needed. Let' see.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Sometimes you want to catch throwable, but in this case it's just work time catct, it will be changed later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Work in progress. Let's se how it ends.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: We have inbound and outbound keys, so we need to handle the integration state by key in some manner. Also, we can add some metadata in operation table. There is different angles in this, performance and maintainability. Anyway, it's probably not as heavy if you think it further. Also, you should optimize for performance and decrease maintainability only if there is a real reason to do it.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayClient.java
solita-alperttiti: 207 is partial success. I haven't check what this means, yet. But let's check this when it's ready.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayService.java
solita-alperttiti: See above.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayService.java
solita-alperttiti: There is todo comment, so it's now just a placeholdet for actual solution.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
jyrkija: I don't think there's any problem as such - the discussion was about whether to change all properties into class format or not. Since they're used in different places, often one at a time, I don't see any benefit in that, but it makes sense in clear contexts like this one, where a group of things needs to be configured (and consumed) all at once. That's how it's done sms & signature configs already.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
jyrkija: Yeah, wip is fine as such, the comments are obviously a bit early :)

However, I really haven't ever encountered a valid case to catch Throwable. If catch-all is needed (and it shouldn't be the go-to) then Exception is usually the correct level. Throwable only adds JVM Errors to the mix, and even by Java documentation the Error class “indicates serious problems that a reasonable application should not try to catch“. It's hard to even make a scenario where catching such issues would run reliably and do something useful. Instead, such a catch can prevent clean shutdowns/kills from going through as intended or cause hangs or delays in shutting down containers.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
jyrkija: Certainly not worth it to early-optimize at the cost of maintainability. I don't think that outbound- and inbound- integrations necessiate state-by-key, though. Main point of my suggestion was that it could be much simpler if we don't maintain any state on key level, but there could be a use for that data as well. Will have to think about it a bit more.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/eu/FederationGatewayClient.java
jyrkija: Sure. And passing the http result is fine if that's needed. I was just musing on the role of this class as an abstraction. If you were to replace the http-request with an ftp fetch, the api could be mostly the same... except the http status code.

exposure-notification/pom.xml
solita-alperttiti: It's needed for ssl. At least for current implementation.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: yep, added this.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Changed this a little bit, so there is no mass update. However, it would be nice to know what is the situation in real life where this is going to be a problem

exposure-notification/pom.xml
solita-alperttiti: Also, version is managed by spring, so no need to define it in here.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+                .build();
+    }
+
+    private KeyStore keyStore(String keyStoreFile, char[] password) throws Exception {
+        File file = ResourceUtils.getFile(keyStoreFile);
+        return loadKeyStore(file, password);
+    }
+
+    private KeyStore loadKeyStore(File file, char[] password) throws Exception {
+        KeyStore keyStore;
```
jyrkija: Keystore variable and return clause might as well be inside the try-block

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+                .setReadTimeout(readTimeout)
+                .setConnectTimeout(socketTimeout)
+                .requestFactory(this::requestFactory)
+                .errorHandler(new RestTemplateErrorHandler())
+                .build();
+    }
+
+    private HttpComponentsClientHttpRequestFactory requestFactory() {
+        try {
+            SSLContext context = (!trustStorePath.isBlank() &&
```
jyrkija, line 63: This looks pretty neat now. 
As tiny prettification, I think it would be possible to trick away this isBlank & length checking here, if you switch the config into a single verified object, like the current SmsConfig and then make that object optional as a whole in this class' constructor. Then you could here use it something like efgsSslConfig.map(this::createSslContext).orElseGet(this::createDefaultSslContext))

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/MaintenanceService.java
```diff
@@ -59,4 +64,24 @@ public void runMaintenance() {
                 keyValue("removedBatches", removedKeys),
                 keyValue("addedBatches", addedBatches));
     }
+
+    // TODO: rename service or make own service for this
```
jyrkija: I think I'd vote for a separate one, but either will do

exposure-notification/src/main/resources/application-dev.yml
```diff
@@ -15,3 +15,6 @@ covid19:
       randomize-key: true
   publish-token:
     url: "${EN_PT_URL:http://localhost:8081}"
+  federation-gateway:
+    client-sha256: "${EN_DEV_EFGS_SHA256:"
```
jyrkija: Ending } is missing

exposure-notification/src/main/resources/application.yml
```diff
@@ -84,3 +84,19 @@ covid19:
     token-verification-lifetime: P14D
   publish-token:
     url: "${EN_PT_URL:}"
+  federation-gateway:
+    url-template: http://localhost:8080/diagnosiskeys/{operation}/{variable}
+    trust-store:
+      path: "${tech-ec-europa-eu:}"
```
jyrkija, line 90: Supposed to come through env, right? For consistency with other env variables, should be upper cased. Also, for other secrets, I've avoided putting any devenv defaults into the application.yml. They can be in the application-dev.yml or (especially for any secrets) just injected through env in the idea launch config.

exposure-notification/src/main/resources/db/migration/V03.01__create_efgs_operation_table.sql
```diff
@@ -0,0 +1,20 @@
+create type en.state_t as enum('QUEUED', 'STARTED', 'FINISHED', 'ERROR');
+create type en.direction_t as enum('INBOUND', 'OUTBOUND');
+
+create table en.efgs_operation (
+    id bigint primary key generated always as identity,
+    state en.state_t not null default 'STARTED',
```
jyrkija: Do you need a default? It might be more explicit to just set it in the insert.

exposure-notification/src/main/resources/db/migration/V03.01__create_efgs_operation_table.sql
```diff
@@ -0,0 +1,20 @@
...
+    direction en.direction_t not null,
+    keys_count_total int,
+    keys_count_201 int,
+    keys_count_409 int,
+    keys_count_500 int,
+    updated_at timestamptz not null default now(),
+    run_count int not null default 0
+);
+
+alter table en.diagnosis_key add efgs_operation bigint;
```
jyrkija, line 16: So all new keys are automatically set to a queued operation and the value is never null, or is null still a possibility?

exposure-notification/src/main/resources/db/migration/V03.01__create_efgs_operation_table.sql
```diff
@@ -0,0 +1,20 @@
...
+alter table en.diagnosis_key add constraint fk_efgs_operation foreign key (efgs_operation) references en.efgs_operation(id);
+insert into en.efgs_operation (state, direction) values ('QUEUED', 'OUTBOUND');
+update en.diagnosis_key set efgs_operation=(select id from en.efgs_operation where state='QUEUED');
```
jyrkija, line 19: The default value for unknown "acceptance for sending keys to EU" is false, so we shouldn't queue all pre-existing keys for sending either.

exposure-notification/src/main/resources/db/migration/V03.02__alter_diagnosis_key.sql
```diff
@@ -0,0 +1,4 @@
+alter table en.diagnosis_key add origin varchar(2) not null default '';
+alter table en.diagnosis_key add visited_countries varchar(2)[] not null default '{}';
+alter table en.diagnosis_key add days_since_onset_of_symptoms int not null default 0;
```
jyrkija: Perhaps we should retain the knowledge of "not known" (null) vs 0 days and default the value in the code, so we can adjust the logic later.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/EfgsServiceWithDaoTestIT.java
```diff
@@ -0,0 +1,220 @@
...
+
+    private HttpHeaders getDownloadResponseHeaders() {
+        HttpHeaders headers = new HttpHeaders();
+        headers.add(NEXT_BATCH_TAG_HEADER, "null");
+        headers.add(BATCH_TAG_HEADER, "test-1");
+        return headers;
+    }
+
+    private void deleteOperations() {
+        String sql = "delete from en.efgs_operation where state <> CAST(:state as en.state_t)";
```
jyrkija: cast in caps

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/EfgsServiceWithDaoTestIT.java
```diff
@@ -0,0 +1,220 @@
...
+                .andRespond(withStatus(HttpStatus.OK)
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .body("")
+                );
+        dd.addKeys(1, md5DigestAsHex("test".getBytes()), to24HourInterval(Instant.now()), keyGenerator.someKeys(5), 5);
+
+        try {
+            federationGatewayService.startOutbound();
+        } catch (Exception e) {
+            assertOperationErrorStateIsCorrect(FederationOperationDao.EfgsOperationDirection.OUTBOUND);
```
jyrkija: I find it easier to read with statically imported enum names where the context makes it obvious what enum it is

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/diagnosiskey/TestKeyGenerator.java
```diff
@@ -22,25 +19,59 @@ public TestKeyGenerator(int seed) {
...
-        return someKeys(count, count);
+        return someKeys(count, count, true);
     }
 
-    public List<TemporaryExposureKey> someKeys(int count, int symptomsDays) {
+    public List<TemporaryExposureKey> someKeys(int count, boolean consentToShare) {
+        return someKeys(count, count, consentToShare);
+    }
+
+    public List<TemporaryExposureKey> someKeys(int count, int symptomsDays, boolean consentToShare) {
```
jyrkija, line 29: Didn't spot this being used anywhere. Would seem worthwhile to confirm that without consent, nothing is sent.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/EfgsServiceWithDaoTestIT.java
```diff
@@ -0,0 +1,220 @@
...
+    public void uploadKeys() {
+        mockServer.expect(ExpectedCount.once(),
+                requestTo("http://localhost:8080/diagnosiskeys/upload/"))
+                .andExpect(method(HttpMethod.POST))
+                .andRespond(withStatus(HttpStatus.OK)
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .body("")
+                );
+        dd.addKeys(1, md5DigestAsHex("test".getBytes()), to24HourInterval(Instant.now()), keyGenerator.someKeys(5), 5);
+        assertUploadOperationStateIsCorrect(federationGatewayService.startOutbound(), 5, 5, 0, 0);
```
jyrkija, line 110: Perhaps this should also verify that the inserted key is tied to the correct outbound operation

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
```diff
@@ -0,0 +1,125 @@
...
+
+import static fi.thl.covid19.exposurenotification.efgs.FederationGatewayBatchUtil.*;
+
+@Service
+public class FederationGatewayService {
+
+    private static final int BATCH_MAX_SIZE = 5000;
+
+    private final FederationGatewayClient client;
+    private final DiagnosisKeyDao dd;
```
jyrkija, line 25: Not a fan of abbreviated names. Had to scroll all the way back up here to see what these were

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
```diff
@@ -0,0 +1,125 @@
...
+    private void doInbound(String date, Optional<String> batchTag) {
+        client.download(date, batchTag).forEach(
+                d -> dd.addInboundKeys(transform(d), IntervalNumber.to24HourInterval(Instant.now()))
+        );
+    }
+
+    private void doOutbound(long operationId) {
+        boolean finished = false;
+        try {
+            List<List<TemporaryExposureKey>> localKeys = partitionIntoBatches(dd.fetchAvailableKeysForEfgs(operationId));
```
jyrkija, line 71: The batching seems kind of pointless when we don't track state for each batch. If one fails (potentially with a connection exception) we mark the whole operation failed and if we want to retry something, we need to re-send all the batches of the operation.

jyrkija: That is: I realize the efgs wants a small batch-size, but I think our operation concept should coincide with batch, not something larger.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+
+    public FederationOperationDao(NamedParameterJdbcTemplate jdbcTemplate) {
+        this.jdbcTemplate = requireNonNull(jdbcTemplate);
+    }
+
+    public long getQueueId() {
+        String sql = "select id from en.efgs_operation where state = CAST(:state as en.state_t) and " +
+                "direction = CAST(:direction as en.direction_t)";
```
jyrkija: caps casts, through the class

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
```diff
@@ -0,0 +1,157 @@
...
+    public long startOutboundOperation() {
+        String lock = "lock table en.efgs_operation in access exclusive mode";
+        jdbcTemplate.update(lock, Map.of());
+        if (isOutboundOperationAvailable()) {
+            long operationId = this.getQueueId();
+            markOutboundOperationStartedAndCreateNewQueue(operationId);
+            return operationId;
+        } else {
+            LOG.info("Update to efgs unavailble. Skipping.");
+            return -1;
```
jyrkija, line 53: Optional is as better "not available" than a magic value. It also makes it that the user can just:
startOuboundOperation().ifPresent(this::doTheThing)

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private boolean isOutboundOperationAvailable() {
+        String sql = "select count(*) from en.efgs_operation " +
+                "where state = CAST(:state as en.state_t) and direction = CAST(:direction as en.direction_t)";
+        return jdbcTemplate.query(sql, Map.of(
+                "state", EfgsOperationState.STARTED.name(),
+                "direction", EfgsOperationDirection.OUTBOUND.name()
+        ), (rs, i) -> rs.getInt(1))
+                .stream().findFirst().orElseThrow(() -> new IllegalStateException("Count returned nothing.")) == 0;
```
jyrkija, line 138: I think there's a bit of awkwardness/complexity due to maintaining the next operation beforehand (all new keys tied to the pre-created upcoming operation). It also means that the operations cannot easily be limited to a specific size (what I commented above on the batches) which kind of ruins the usability of the results tracking. 
I know I critized myself about updating all rows for the operation, but this isn't quite what I meant either. I think the previous method of just updating the rows is better, if we really want to have the key->operation linking in this direction.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationSyncProcessor.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+
+@Component
+public class FederationSyncProcessor {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FederationSyncProcessor.class);
+
+    private final FederationGatewayService fgs;
```
jyrkija, line 15: just gatewayService or even federationGatewayService would be a clearer name

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/error/EfgsOperationException.java
```diff
@@ -0,0 +1,12 @@
+package fi.thl.covid19.exposurenotification.error;
+
+public class EfgsOperationException extends RuntimeException {
```
jyrkija, line 3: Didn't see this caught anywhere or handled differently. What's the special exception for?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/UploadResponseEntity.java
```diff
@@ -0,0 +1,7 @@
+package fi.thl.covid19.exposurenotification.efgs;
+
+import java.util.HashMap;
+import java.util.List;
+
+public class UploadResponseEntity extends HashMap<Integer, List<Integer>> {
```
jyrkija, line 6: I kind of like that this has a name for our code, but it can contain nulls, unlike anything else, so it can be misleading. I also don't think that there's much to be done about it - the server might return nulls in any event. Perhaps it would be better to make this class a private internal one to the service that uses it.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
```diff
@@ -0,0 +1,157 @@
...
+                "state", FederationOperationDao.EfgsOperationState.QUEUED.name(),
+                "direction", FederationOperationDao.EfgsOperationDirection.OUTBOUND.name()
+        ), Long.class));
+    }
+
+    @Transactional(timeout = 10)
+    public long startOutboundOperation() {
+        String lock = "lock table en.efgs_operation in access exclusive mode";
+        jdbcTemplate.update(lock, Map.of());
+        if (isOutboundOperationAvailable()) {
```
jyrkija, line 47: Rather than 3 separate queries (isAvailable, getId, markId), I htink the markOutboundStarted... could just use the postgresql "returning id" syntax do the check and update in one go. Then this whole if-structure could become something like return startNextOperation(), returning an Optional operation id

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -56,7 +56,7 @@ public int deleteVerificationsBefore(Instant verificationTime) {
     @Transactional
     public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys, long exportedKeyCount) {
         if (verify(verificationId, requestChecksum, keys.size(), exportedKeyCount) && !keys.isEmpty()) {
-            batchInsert(interval, keys);
+            batchInsert(interval, keys, federationOperationDao.getQueueId());
```
jyrkija, line 59: This will read the queue id from DB and then do the insert in a parallel transaction. While it's still building the batch to insert, another thread can start the upload operation for that queue id and fetch the keys to send. As a result, these keys might not be in the batch because the sender doesn't see them yet

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -118,24 +119,57 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
+        String sql = "select key_data, rolling_period, rolling_start_interval_number, transmission_risk_level, " +
+                "visited_countries, days_since_onset_of_symptoms, origin, consent_to_share " +
+                "from en.diagnosis_key " +
+                "where efgs_operation = :efgs_operation " +
+                "order by key_data";
+
+        return new ArrayList<>(jdbcTemplate.query(sql, Map.of("efgs_operation", operationId), (rs, i) -> mapKey(rs)));
+    }
+
+
```
jyrkija: extra newlines

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
jyrkija: Replying to myself: no that cannot happen. The lock on operation start is access exclusive so it will wait for this op (or the other way around). Still, I'd prefer a solution that is a bit more obvious and doesn't need this diagnosisdao -> federationdao dependency.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyService.java
```diff
@@ -36,45 +38,44 @@ public DiagnosisKeyService(DiagnosisKeyDao dao, PublishTokenVerificationService
         LOG.info("Initialized");
     }
 
-    public void handlePublishRequest(String publishToken, List<TemporaryExposureKey> keys) {
+    public void handlePublishRequest(String publishToken, List<TemporaryExposureKeyRequest> requestKeys) {
         Instant now = Instant.now();
-        List<TemporaryExposureKey> filtered = filter(keys, now);
         PublishTokenVerification verification = tokenVerificationService.getVerification(publishToken);
+        List<TemporaryExposureKey> keys = transform(requestKeys, verification.symptomsOnset);
```
jyrkija: This switches the ordering of transform/filter and the new transform validates that the time of the key is within 14 days (and throws an exception if it's not). We can get keys that are in the past in a valid request: those are filtered here, so the transform should happen after the filter (or be less rigid in the validation).

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/Validation.java
```diff
@@ -66,6 +72,28 @@ public static int validateRollingPeriod(int rollingPeriod) {
...
+        }
+    }
+
+    public static boolean validateISOCountryCode(String code) {
+            return Locale.getISOCountries(Locale.IsoCountryCode.PART1_ALPHA2).contains(code);
+    }
+
+    public static int validateDaysSinceOnsetOfSymptoms(int daysSinceOnsetOfSymptoms) {
+        return verifyValueBetween(
+                "days since onset of symptoms", daysSinceOnsetOfSymptoms,
```
jyrkija, line 93: Does efgs actually promise this to be +-14? It's an ok logic for us, but if it's not guaranteed by them, we might just want to filter them, rather than validate. If it's promised by the API though, then this is fine.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
...
+                .addRDN(BCStyle.C, "FI")
+                .addRDN(BCStyle.CN, "koronavilkku.dev")
+                .addRDN(BCStyle.O, "koronavilkku dev")
+                .build();
+
+        X509v3CertificateBuilder certBuilder = new X509v3CertificateBuilder(
+                subject,
+                BigInteger.valueOf(123456789),
+                Date.from(Instant.now()),
+                Date.from(Instant.now().plus(Duration.ofDays(100))),
```
jyrkija, line 148: Call Instant.now() only once. It's not likely the time changes much between those lines, but...

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
+package fi.thl.covid19.exposurenotification.efgs;
+
+import com.google.protobuf.ByteString;
+import com.google.protobuf.ProtocolStringList;
+import fi.thl.covid19.proto.EfgsProto;
+import org.bouncycastle.asn1.x500.X500Name;
```
jyrkija, line 6: I saw a lot of bouncycastle in various examples, but found that it didn't do anything for me in addition to java.security packages when I wrote the original batch signing. I guess there was something specific about the cert formats here so you couldn't just use/extend the current Signing-class logic for here?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
...
+import java.security.cert.*;
+import java.security.cert.Certificate;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Base64;
+import java.util.Comparator;
+import java.util.Date;
+
+@Component
+public class FederationBatchSigner {
```
jyrkija, line 42: This class probably merits it's own test set as well.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
...
+                SubjectPublicKeyInfo.getInstance(keyPair.getPublic().getEncoded()));
+        JcaContentSignerBuilder builder = new JcaContentSignerBuilder(DIGEST_ALGORITHM + "RSA");
+        ContentSigner signer = builder.build(keyPair.getPrivate());
+
+        byte[] certBytes = certBuilder.build(signer).getEncoded();
+        CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+        return (X509Certificate) certificateFactory.generateCertificate(new ByteArrayInputStream(certBytes));
+    }
+
+    private byte[] generateBytesForSigning(EfgsProto.DiagnosisKeyBatch batch) {
```
jyrkija, line 159: Stylistical stuff: I'd prefer taking in the smallest possible argument for easier testability and reuse. This function doesn't seem to need a batch, it just needs the keys. In fact, this whole class doesn't seem to touch anything else from the batch.

Also, the classname "...Signer" doesn't make me assume that it will "generate bytes to verify" - more a job for whomever call the signing function. Is the signing data actually a different serialization of the batch keylist?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
...
+        writeIntInByteArray(diagnosisKey.getDaysSinceOnsetOfSymptoms(), byteArrayOutputStream);
+        writeSeperatorInArray(byteArrayOutputStream);
+        return byteArrayOutputStream.toByteArray();
+    }
+
+    private String bytesToBase64(byte[] bytes) {
+        try {
+            return Base64.getEncoder().encodeToString(bytes);
+        } catch (IllegalArgumentException e) {
+            LOG.error("Failed to convert byte array to string");
```
jyrkija, line 198: Why do you want to catch this? If the encoding fails, what's the point of continuing?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
...
+    ) throws Exception {
+        this.keyStorePath = keyStorePath;
+        this.keyStorePassword = keyStorePassword.toCharArray();
+        this.keyStoreKeyAlias = keyStoreKeyAlias;
+        this.keyStore = initKeystore();
+    }
+
+    public String sign(final EfgsProto.DiagnosisKeyBatch data) {
+        try {
+            return sign(data, keyStore);
```
jyrkija, line 66: If the keys were in the caller service, this whole class could be a static util.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayBatchUtil.java
```diff
@@ -0,0 +1,82 @@
...
+            throw new IllegalStateException("Incorrect format from federation gateway.", e);
+        }
+    }
+
+    public static String getBatchTag(Instant instant, String postfix) {
+        return getDateString(instant) + "-" + postfix;
+    }
+
+    public static String getDateString(Instant date) {
+        return DateTimeFormatter.ISO_DATE.format(LocalDate.ofInstant(date, ZoneId.of("UTC")));
```
jyrkija: You can use ZoneOffset.UTC instead of fetching the zoneid with String.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
```diff
@@ -0,0 +1,239 @@
...
+        X500Name subject = new X500NameBuilder(BCStyle.INSTANCE)
+                .addRDN(BCStyle.C, "FI")
+                .addRDN(BCStyle.CN, "koronavilkku.dev")
+                .addRDN(BCStyle.O, "koronavilkku dev")
+                .build();
+
+        X509v3CertificateBuilder certBuilder = new X509v3CertificateBuilder(
+                subject,
+                BigInteger.valueOf(123456789),
+                Date.from(Instant.now()),
```
jyrkija, line 147: They don't have an instant-based API? No-one should be using java.util.Date any more.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
```diff
@@ -0,0 +1,125 @@
...
+        long operationId = fod.startOutboundOperation();
+
+        if (operationId > 0) {
+            doOutbound(operationId);
+        }
+        return operationId;
+    }
+
+    public void startInbound(Optional<String> batchTag) {
+        String date = getDateString(Instant.now());
```
jyrkija, line 51: If inbound ops trigger on both nodes at the same time, both fetch the same data, right? It's probably not a big issue, but we could use the status-table to only have one of these running at a time.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
```diff
@@ -0,0 +1,125 @@
...
+        doInbound(date, batchTag);
+    }
+
+    public void startErronHandling() {
+        fod.setStalledToError();
+        List<Long> errorOperations = fod.getOutboundOperationsInError();
+        errorOperations.forEach(this::doOutbound);
+    }
+
+    // TODO: which interval should be used? Should we change interval to be more precise?
```
jyrkija, line 61: The interval-number is used to determine which out-going batch the data ends up in. It's semantic is "when we received the data". So I guess this is fine.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -118,24 +119,57 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
+
+    @Transactional
+    public void addInboundKeys(List<TemporaryExposureKey> keys, int interval) {
+        if (!keys.isEmpty()) {
+            boolean finished = false;
+            long operationId = federationOperationDao.startInboundOperation();
```
jyrkija, line 137: I think it would be cleaner if these calls came from the service doing the sync, not this dao. That keeps the dao's less connected and makes it possible to implement "fetch only once" despite multiple nodes. Of course, it needs more logic for timeout and such as well - this method does a single transaction so there's no middle states.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: I don't understand. I think @Transactional timeout with jdbctemplate sets timeout at jdbc statement level before each statement, so if app is working this should be enough. Or even if it's shut-down without total crash. However, in case of crash in the middle of transaction it's probably more complicated thing. Then it's most likely about when the underlying tcp connection to db get killed and after that db will do rollback and releases the lock on the same time. Problematic situation would be the case where connection stays alive even the app has been crashed on the right moment. We might send statement_timeout before actual statement to db to be sure, but let's think this through with patience.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayBatchUtil.java
solita-alperttiti: Yep, true.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/MaintenanceService.java
solita-alperttiti: This was already done.

exposure-notification/src/main/resources/application.yml
solita-alperttiti: I got these from kela. Need to check these anyway when we're going to deploy this. So, for now these stays like that.

exposure-notification/src/main/resources/db/migration/V03.01__create_efgs_operation_table.sql
solita-alperttiti: Yep, that is probably true. I had some nice idea with this, but yes, I'll change that.

exposure-notification/src/main/resources/db/migration/V03.01__create_efgs_operation_table.sql
solita-alperttiti: Yep, with current implementation it is not possible after initial update for old rows. I'll change that.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
jyrkija: What the...? This comment was written ages ago, and on a very different version of this file. It must have been left-over from some uncommitted review and pulled into this one. I didn't write it today. Just ignore it, I'll check the retry-logic from the current code and comment if needed.

exposure-notification/src/main/resources/db/migration/V03.01__create_efgs_operation_table.sql
solita-alperttiti: This is not decided yet. Anyway, while this seems a little bit awkward, we'll need initial queue and this way these old keys are also retrofitted into the new pattern. Because consent is set to false, those are not send.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
jyrkija: I checked how the mid-operation failures work in the current implementation and I don't see any issue with that.

exposure-notification/src/main/resources/db/migration/V03.02__alter_diagnosis_key.sql
solita-alperttiti: We already have decision that default will be 6. Anyway, it's still might be a good thing to do in a way you suggest. I'll change that in pr where DOSS logic will be implemented.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/diagnosiskey/TestKeyGenerator.java
solita-alperttiti: It will be, it will be.. :)

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/EfgsServiceWithDaoTestIT.java
solita-alperttiti: Added, todo for this for now. More tests are needed.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
solita-alperttiti: Yep, I'll check this on next refactoring round.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
solita-alperttiti: Yep, that's correct and it's because efgs we''l do this batching. Also, this needs some changes to our concept, so for now it is in that way. This will be resolved later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
solita-alperttiti: Yep, probably true. Anyway, I'll leave that for next refactoring phase.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
solita-alperttiti: Yep, good points. Anyway,  let's figure this out later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationSyncProcessor.java
solita-alperttiti: I dont't like long names. If you work with code, it's quite clear what it means. Anyway, I'll keep that in mind on next refactoring round.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/error/EfgsOperationException.java
solita-alperttiti: Currently none. There was some idea, but it might be already dropped. I'll fix that later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/UploadResponseEntity.java
solita-alperttiti: Yep, I'll put that on refactoring list.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
solita-alperttiti: Might be. However, then we just have one very specific operation and not so generic pieces. I'll let this be for now.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, let's get back into this later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/Validation.java
solita-alperttiti: No, it's more complicated, but this work is on progress and will be implemented on new pr.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: It's start and expiration date for randomly generated dev cert, which is there just that project will be run without actual certs, yes, it's kind of cleaner, but it doesn't make much difference if the expiration is millisecond or two later tha 100 days.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: I have used logic from efgs and it's uses bouncy castle. So, it's easier to check that our signing is compatible with efgs and also there is some more intuitive top-level methods in bouncy castle. Anyway, it could be done with java also.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: Yep.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: This is more like an encapsulation of all of this signing thing. However naming could be also something else. There is some logic straight from the efgs. It's simpler to use same approach to make our implementation compatible with it, even if there is some stylistic things.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: This is from efgs. It fails on the other end in a same way.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: could be, see from the other comments.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: dunno... this is latest I think.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
solita-alperttiti: This is a good point, however efgs maintains state for these, so this probably isn't a problem. But I haven't checked this yet in this level. However, with callback we are always on one node. This is a potential problem only with scheduled run.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, I'll leave that to next refactoring.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
solita-alperttiti: Yep, good point. I'll check this later.

solita-alperttiti: Actually, I checked this briefly and I don't quite get what you mean by making whole class optional. There is 4 values and yes, those can be bundled into configuration properties, but don't get how this solves anything. You could move the logic inside properties yes, but not so sure if this is really much better. Maybe I just don't get what the actual point in here is.

solita-alperttiti: Anyway, I moved these to properties into property class and moved validation logic also inside that class. It's at least a little bit cleaner now.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
solita-alperttiti: Also, I think we're not gonna hit this limit very soon, even if we upload only once in a day. So, while point is not going anywhere, in practice this is just a failsafe and we should not have need to split our upload anytime soon. So, I'll get back into this later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationOperationDao.java
solita-alperttiti: Fixed in new branch

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/UploadResponseEntity.java
solita-alperttiti: This is actually workaround for resttemplate api to define generics and it's used in client and service classes. Anyway, I'll look into this later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationBatchSigner.java
solita-alperttiti: Anyway, we probably could do this diffrently, but I like to keep this similar as possible.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Let's see how this ends. I'll keep this in mind when there is refactoring time.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientConfiguration.java
jyrkija: The point was that since you're using it as all-or-nothing, you can bundle the properties up and validate that each field inside exists and is good - the whole thing must be missing if anything is. Then in here, you only need to check whether you have a config or don't - it can never be invalid/partial so you don't need to check each field for length/blank (mid-states are in.

The Optional part is just styling on top of that. You can take the trustStorePath as an argument for createSslContext() rather than using it from the field (more explicit) and then this becomes a question of:
`context = config != null ? createSslContext(config) : SSLContextBuilder.create().build()` 
vs 
`config.map(this::createSslContext).orElseGet(this::createDefaultSslContext)`

solita-alperttiti: Are you sure that config can be missing. I think Spring Boot application won't start if configuration is not initialized properly. Anyway, I moved initialization state in the own method inside configuration properties.

jyrkija: Not 100% as I haven't done it on a configuration bean, but I think so. There's a variety of ways to make a beans optional (conditionalonproperty for example), and I don't see why they wouldn't work on configuration beans as well. You can certainly take a dependency-bean in as optional.
That aside, your solution looks good too.

jyrkija: According to docs, conditional configuration beans are a thing:
https://docs.spring.io/spring-boot/docs/1.2.1.RELEASE/reference/html/boot-features-developing-auto-configuration.html

solita-alperttiti: yep, only problem might be if conditional annotations can be used in ConfigurationProperties. I tried this briefly, with following annotations in property class:

`@ConditionalOnExpression("${covid19.federation-gateway.rest-client.trust-store.path} > ''")
@ConditionalOnProperty("${foo.bar} == true")`

Separately, not the same time. I expect to get null pointer exception with both, but everything runs smoothly, which indicates that conditional have no effect on property class. Anyway, this needs more investigation, maybe later .. :)

solita-alperttiti: Anyway, even if this approach don't work directly in properties, this strategy could be used. However, result might not be as simple as one might think on first look.

PR #34, opened 2020-10-20
Collect key counts per publish operation for statistics

solita-alperttiti:
This pull request resolves key counts for statistics.

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
@@ -54,8 +54,8 @@ public class DiagnosisKeyDao {
     }
 
     @Transactional
-    public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys) {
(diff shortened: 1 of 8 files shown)


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -54,8 +54,8 @@ public int deleteVerificationsBefore(Instant verificationTime) {
     }
 
     @Transactional
-    public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys) {
-        if (verify(verificationId, requestChecksum) && !keys.isEmpty()) {
+    public void addKeys(int verificationId, String requestChecksum, int interval, List<TemporaryExposureKey> keys, long exportedKeyCount) {
```
jyrkija, line 57: The filtering is in the DAO, so I think the exportedCount calculation might as well be also. That would avoid making so many test-changes since you wouldn't need to change the function args.

jyrkija (approved):
Looks good. I wouldn't see it as bad to count in the DAO to avoid the change escalating, but this puts the application logic in the service which is nice.

PR #35, opened 2020-11-02
Parallel bucket calculation

solita-alperttiti:
This pull request resolves COVID-444 .

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/configuration/ConfigurationDao.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/configuration/ConfigurationDao.java
@@ -37,7 +37,9 @@ public class ConfigurationDao {
                 "days_since_last_exposure_scores, " +
                 "duration_scores, " +
                 "transmission_risk_scores, " +
-                "duration_at_attenuation_thresholds " +
+                "duration_at_attenuation_thresholds, " +
+                "duration_at_attenuation_weights, " +
+                "exposure_risk_duration " +
                 "from en.exposure_configuration " +
                 "order by version desc " +
                 "limit 1";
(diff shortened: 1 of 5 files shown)


nmurtola (commented):
Looks good

spheroid:
The configuration parameters are correct.

exposure-notification/src/main/resources/db/migration/V02.03__alter_configuration_table.sql
```diff
@@ -0,0 +1,2 @@
+alter table en.exposure_configuration add duration_at_attenuation_weights decimal(3,2) array[3];
```
jyrkija: How about nulls? I think we should either handle nulls in the fetch-part or add the columns as non-null with default.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/configuration/ConfigurationDao.java
```diff
@@ -48,11 +50,14 @@ public ExposureConfiguration getLatestExposureConfiguration() {
                 toList(rs.getArray("days_since_last_exposure_scores")),
                 toList(rs.getArray("duration_scores")),
                 toList(rs.getArray("transmission_risk_scores")),
-                toList(rs.getArray("duration_at_attenuation_thresholds"))
+                toList(rs.getArray("duration_at_attenuation_thresholds")),
+                toList(rs.getArray("duration_at_attenuation_weights")),
+                rs.getInt("exposure_risk_duration")
```
jyrkija, line 55: I still don't think this name is very descriptive, but we had that discussion elsewhere.

jyrkija (changes requested):

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/configuration/ConfigurationDao.java
solita-alperttiti: Yep, it might be something else or more descriptive e.g. time unit included in name, but we decided to keep it as it is and we could add some documentation if necessary.

exposure-notification/src/main/resources/db/migration/V02.03__alter_configuration_table.sql
solita-alperttiti: yep, this is a good point and I actually just bypassed this thing and then forgot it. I'll add not null into these.

jyrkija (approved):

PR #36, opened 2020-11-11
Dsos mapping

solita-alperttiti:

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
@@ -1,0 +1,74 @@
+package fi.thl.covid19.exposurenotification;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.ConstructorBinding;
+
+@ConfigurationProperties(prefix = "covid19.federation-gateway.rest-client")
+@ConstructorBinding
+public class FederationGatewayRestClientProperties {
+
+    private final TrustStore trustStore;
+    private final ClientKeyStore clientKeyStore;
+
+    public FederationGatewayRestClientProperties(TrustStore trustStore, ClientKeyStore clientKeyStore) {
+        this.trustStore = trustStore;
+        this.clientKeyStore = clientKeyStore;
+    }
+
+    public boolean isMandatoryPropertiesAvailable() {
+        return !this.getTrustStore().path.isBlank() &&
+                this.getTrustStore().password.length > 0 &&
+                !this.clientKeyStore.getPath().isBlank() &&
+                this.clientKeyStore.getPassword().length > 0;
+    }
+
+    public static class TrustStore {
+        private final String path;
+        private final char[] password;
+
+        public TrustStore(String path, char[] password) {
+            this.path = path;
+            this.password = password;
+        }
+
+        public String getPath() {
+            return path;
+        }
+
+        public char[] getPassword() {
+            return password;
+        }
+    }
+
+    public static class ClientKeyStore {
+        private final String path;
+        private final char[] password;
+        private final String alias;
+
+        public ClientKeyStore(String path, char[] password, String alias) {
+            this.path = path;
+            this.password = password;
+            this.alias = alias;
+        }
+
+        public String getPath() {
+            return path;
+        }
+
+        public char[] getPassword() {
+            return password;
+        }
+
+        public String getAlias() {
+            return alias;
+        }
+    }
+
+    public TrustStore getTrustStore() {
+        return trustStore;
+    }
+
+    public ClientKeyStore getClientKeyStore() {
+        return clientKeyStore;
+    }
+}
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
@@ -7,8 +7,8 @@ import org.springframework.boot.context.properties.ConstructorBinding;
 @ConstructorBinding
 public class FederationGatewayRestClientProperties {
 
-    private final TrustStore trustStore;
-    private final ClientKeyStore clientKeyStore;
+    public final TrustStore trustStore;
+    public final ClientKeyStore clientKeyStore;
 
     public FederationGatewayRestClientProperties(TrustStore trustStore, ClientKeyStore clientKeyStore) {
         this.trustStore = trustStore;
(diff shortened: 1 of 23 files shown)


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
```diff
@@ -0,0 +1,74 @@
+package fi.thl.covid19.exposurenotification;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.ConstructorBinding;
+
+@ConfigurationProperties(prefix = "covid19.federation-gateway.rest-client")
+@ConstructorBinding
+public class FederationGatewayRestClientProperties {
+
+    private final TrustStore trustStore;
```
jyrkija: The object is immutable, which is good. With that, I don't see any benefit to having the fields private with getters; public final is just as good and drops out the unnecessary boilerplate.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -191,24 +191,25 @@ private TemporaryExposureKey mapKey(ResultSet rs) throws SQLException {
...
-                rs.getInt("days_since_onset_of_symptoms"),
+                Optional.ofNullable((Integer) rs.getObject("days_since_onset_of_symptoms")),
                 rs.getString("origin"),
                 rs.getBoolean("consent_to_share")
         );
     }
 
     private Map<String, Object> createParamsMap(int interval, TemporaryExposureKey key, long operationId) {
-        return Map.of(
+        Map<String, Object> params = new HashMap<>(Map.of(
```
jyrkija: Ah, nulls in maps. I really wish jdbctemplate would take in Optional as value, but alas... 
No pretty way around it, but instead of creating 2 maps, might as well just make an empty hashmap and put in all the values.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyService.java
```diff
@@ -65,7 +67,11 @@ private long getExportedKeyCount(List<TemporaryExposureKey> keys) {
                 requestKey.rollingStartIntervalNumber,
                 requestKey.rollingPeriod,
                 requestKey.visitedCountries,
-                Math.toIntExact(ChronoUnit.DAYS.between(symptomsOnset, utcDateOf10MinInterval(requestKey.rollingStartIntervalNumber))),
```
jyrkija, line 68: Using the efgs mapper here seems a bit odd, but it's an identity mapping anyhow, right? If it's used for non-efgs keys as well, then maybe it should just be in the diagnosis key package.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/DsosMapperUtil.java
```diff
@@ -0,0 +1,41 @@
...
+            this.mapper = mapper;
+        }
+
+        public Optional<Integer> apply(Integer dsos) {
+            return Optional.ofNullable(mapper.apply(dsos));
+        }
+
+        public static Optional<Integer> mapFrom(int dsos) {
+           DsosInterpretationMapper mapperInRange = Arrays.stream(values())
+                   .filter(e -> e.range.get().anyMatch(i -> i == dsos))
```
jyrkija: Since this is what the range is used for, IntStream is kind of wasteful. The SYMPTOMS_WITH_TIME_RANGE one does 1800 comparisons to know if an int is between 2 values. Maybe just a Predicate<Integer> that just checks against the edges instead. I guess you can even wrap it as a Range class if you want (shame that Java doesn't have a default one).

jyrkija (changes requested):

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/DsosMapperUtil.java
solita-alperttiti: Yep, this is indeed stupid thing. I'm changing that.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyService.java
solita-alperttiti: Yep, idea was to use same thing everywhere, e.g. in case we'll need to map our values to something else, but this is wrong place. I'll remove that.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, could be in that way too. There is no much difference in my opinion, but might be a little bit more uniform.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
solita-alperttiti: Yep, it's cleaner like this.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
```diff
@@ -0,0 +1,46 @@
...
+
+@ConfigurationProperties(prefix = "covid19.federation-gateway.rest-client")
+@ConstructorBinding
+public class FederationGatewayRestClientProperties {
+
+    public final TrustStore trustStore;
+    public final ClientKeyStore clientKeyStore;
+
+    public FederationGatewayRestClientProperties(TrustStore trustStore, ClientKeyStore clientKeyStore) {
+        this.trustStore = trustStore;
```
jyrkija, line 14: Oh, one more thing: In dataobjects, use Objects.requireNonNull (I usually static import it) to make sure they can't be built wrong. Catches any mistakes as soon as possible.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceWithDaoTestIT.java
```diff
@@ -106,8 +140,8 @@ public void uploadKeys() {
                         .body("")
                 );
         dd.addKeys(1, md5DigestAsHex("test".getBytes()), to24HourInterval(Instant.now()), keyGenerator.someKeys(5), 5);
-        // TODO: check keys are bound to correct operation
```
jyrkija, line 109: This verifies that the started operation id matches the one that was queued before. If you want to verify that the keys are bound to that operation, you could grab them from that keyGenerator.someKeys -call and run a separate sql after the insert to check that their operation id matches it also.

jyrkija (approved):

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
solita-alperttiti: I'm not sure if it's necessary for objects managed by spring. It should take care of it and using it feels unnecessary. On every other object I share your opinion. Also, we could move to kotlin. :)

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceWithDaoTestIT.java
solita-alperttiti: Yep, this is true. However, keys and operations are cleared before test, so probably just checking that all keys have correct operation id would do. Dunno.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
jyrkija: Are you sure spring ensures it? I thought missing/null properties are still possible in configuration beans, but if they're not then sure, it's not necessary.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceWithDaoTestIT.java
jyrkija: Yeah, that's valid too

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
solita-alperttiti: I was, but I was wrong. :) In context of configuration properties missing properties will be initialized as null. Spring-way to achieve "not null" seems to be using Validator annotations for fields. This is quite odd that they have class level things like fail on unknown properties and things like that, but not null needs to be done field by field. So, I'll add these in some form. However, for normal beans it shouldn't be possible (if there is not a bug or something), or at least I can't easily imagine in what kind of situation this could be possible.

jyrkija: Yeah, that is a bit odd. That's kind of why I mistrust all lib/framework code. For our own codebase, there are no nulls and that can be trusted. For everyone else's, it's easier to just verify everything "at the border" than dig under the hood every time to check. 

Also, I agree that we could move to kotlin - alas that was not my choice to make :)

PR #37, opened 2020-11-13
CallbackController

solita-alperttiti:


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
```diff
@@ -27,8 +29,8 @@ public boolean isMandatoryPropertiesAvailable() {
         public final char[] password;
 
         public TrustStore(String path, char[] password) {
-            this.path = path;
-            this.password = password;
+            this.path = requireNonNull(path);
```
jyrkija, line 32: The truststore/keystore objects can actually be null also, if you skip the entire segment in the yml.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
```diff
@@ -23,14 +25,14 @@
 
     private final FederationGatewayClient client;
     private final DiagnosisKeyDao dd;
```
jyrkija, line 27: I know I mentioned it before but I have to mention it again; abbreviating variable names is an anti-pattern, unless the abbreviation is clearer than the actual or an established name in itself. All code is clear to the writer, but the point is to be clear to someone else. Naming is there to communicate meaning and names like dd and fod communicate virtually nothing. IDEs are good with autocomplete so writing a proper name shouldn't be that much of a chore.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
```diff
@@ -33,7 +35,7 @@ private void runExportToEfgs() {
             fixedRateString = "${covid19.federation-gateway.download-interval}")
     private void runImportFromEfgs() {
         LOG.info("Starting scheduled import from efgs.");
```
jyrkija, line 37: Might be useful to include in the logs some identifying information for the operation - the date/operation id, etc.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/CallbackController.java
```diff
@@ -0,0 +1,34 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(CallbackController.class);
+
+    private final FederationGatewayService federationGatewayService;
+
+    public CallbackController(FederationGatewayService federationGatewayService) {
+        this.federationGatewayService = federationGatewayService;
+    }
+
+    @GetMapping("/callback")
+    public void triggerCallback(@RequestParam("batchTag") String batchTag, @RequestParam("date") @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) LocalDate date) {
```
jyrkija, line 29: I find it odd that they decided to go with no-zone local dates on an international API... but I guess it is what it is. Is it a documented thing at least that it's interpreted as UTC?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/FederationGatewayRestClientProperties.java
solita-alperttiti: Yep, you are correct. I'll fix that.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
solita-alperttiti: Yep, this is on my not so urgent list. Not sure about anti-pattern though.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
solita-alperttiti: There is a timestamp for log row, not sure what else we could have in here, because this is scheduled operation.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/CallbackController.java
solita-alperttiti: From code:

`.queryParam("date", callbackTask.getBatch().getCreatedAt()      .withZoneSameInstant(ZoneOffset.UTC).format(DateTimeFormatter.ISO_LOCAL_DATE))`

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
solita-alperttiti: E.g. in idea you could press ctrl to see the type of object if you are not sure. Too long names are just hard to read and makes code more messy.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
jyrkija: I was thinking about the date that is given into the startInbound-call (you're right, it's already in the timestamp, so kind of redundant) and maybe the operation id on the finished-log-row (if startInbound returns it). Perhaps it's already obvious from the other rows that get logged during the op.

solita-alperttiti: Yep, currently there is no id returned, but it might be indeed a good addition.  Good point. Let's see after initial testing.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewayService.java
jyrkija: That applies in idea, but not when reviewing the code in github. It also doesn't apply to variables as a whole - only to singletons where the class and the role of the variable are 1-to-1. Often, repeating the classname isn't a very good name, but for these singletons it's also the role of the instance, so it works.

I wholeheartedly disagree that long names are hard to read and messy, unless by "too long" you mean really long names. No-one wants to see 80 char names and yes, shorter is better, but only if the shorter one communicates well. Abbreviations are not easier to read - they are mental mappings to the concept behind them, forcing the reader to do an extra cognitive step as he reads. Clear, descriptive names read naturally and eradicate the need to do that mapping.  Obviously, that's less of an issue when the mapped meaning is right next to the use, (local variable, loops, lambdas) but fields especially can live pretty far from the use-place and when you look at a diff, the declaration might not even be included.

When writing the code, you have it all in your mind anyhow, but a potential reader does not. He likely comes through multiple files/classes/functions on his way to understand some whole and doesn't know the abbreviations of this particular function by heart. On his way, he'll have to do an extra step to check the meaning of each variable, distracting from the real issue each time.

solita-alperttiti: Yep, I agree you have your point. I just have seen also these very long and hard to read namings. I'll make names more understandable after initial testings.

PR #40, opened 2020-11-14
Changed abbreviations to more meaningful form

solita-alperttiti:
For real, this pull request changes key collecting logic into more robust shape, so there is no need for explicit db locks and such. There is also some other tunings for null safety and also db connection query has been removed as an overkill in pool settings (this needs to be verified that there is no some special reason to use it).


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -119,31 +122,59 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
+        List<TemporaryExposureKey> keys = new ArrayList<>(jdbcTemplate.query(sql, Map.of(
+                "min_retry_count", retry ? 1 : 0,
+                "max_retry_count", retry ? MAX_RETRY_COUNT : 1,
+                "timestamp", timestamp
+        ), (rs, i) -> mapKey(rs)));
+
+        if (keys.isEmpty()) {
+            return Optional.empty();
+        } else {
+            return Optional.of(new FederationOutboundOperation(keys, operationDao.startOperation(OperationDao.EfgsOperationDirection.OUTBOUND, timestamp)));
```
jyrkija: The line flows out of view - maybe break it down some. Also you could static import at least the efgsopeartiondirection enum.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -119,31 +122,59 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
         LOG.info("Fetching queued keys not sent to efgs.");
-        String sql = "select key_data, rolling_period, rolling_start_interval_number, transmission_risk_level, " +
-                "visited_countries, days_since_onset_of_symptoms, origin, consent_to_share " +
+        Timestamp timestamp = new Timestamp(Instant.now().toEpochMilli());
+        String sql = "with batch as ( " +
+                "select key_data " +
                 "from en.diagnosis_key " +
-                "where efgs_operation = :efgs_operation " +
-                "order by key_data";
+                "where sent_to_efgs is null and retry_count >= :min_retry_count and retry_count < :max_retry_count limit 5000 ) " +
```
jyrkija: Spent a moment figuring out what does min_retry_count mean, but got it eventually. Perhaps not the most obvious logic at a glance, but I don't have a  clearer suggestion off the top of my hat.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -119,31 +122,59 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
     }
 
-    public List<TemporaryExposureKey> fetchAvailableKeysForEfgs(long operationId) {
+    @Transactional
+    public Optional<FederationOutboundOperation> fetchAvailableKeysForEfgs(boolean retry) {
         LOG.info("Fetching queued keys not sent to efgs.");
-        String sql = "select key_data, rolling_period, rolling_start_interval_number, transmission_risk_level, " +
-                "visited_countries, days_since_onset_of_symptoms, origin, consent_to_share " +
+        Timestamp timestamp = new Timestamp(Instant.now().toEpochMilli());
+        String sql = "with batch as ( " +
```
jyrkija, line 131: This is an interesting query from parallel executions' point of view. Not sure if there is a problem, but let's hash it out in chat.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -119,31 +122,59 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
+            return Optional.empty();
+        } else {
+            return Optional.of(new FederationOutboundOperation(keys, operationDao.startOperation(OperationDao.EfgsOperationDirection.OUTBOUND, timestamp)));
+        }
+    }
 
-        return new ArrayList<>(jdbcTemplate.query(sql, Map.of("efgs_operation", operationId), (rs, i) -> mapKey(rs)));
+    @Transactional
+    public void setNotSent(FederationOutboundOperation operation) {
+        String sql = "update en.diagnosis_key set sent_to_efgs = null where key_data in (:key_datas)";
```
jyrkija: How does this perform when there are thousands of keys in that key set? 
1. I recall from sql server, that this kind of stuff got expanded by jdbc into a long list like "key_data in (?, ?, ?, ?, ...)" which means that each variant of key-count gets a different sql and it won't be precompiled. Not sure if jdbc4 is smarter about that.
2. This has to compare each key-row with the entire list, so if we have a million rows in DB and 5000 rows in the list, it could get slow. If you turn this into a batch update, it should just be 5000 updates with primary key. Again, not sure if there is something that optimizes it better than I assume, but it looks like it could get costly.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -119,31 +122,59 @@ public boolean verify(int verificationId, String requestChecksum, long totalKeyC
...
+        operationDao.markErrorOperation(operation.operationId, Optional.of(operation.batchTag));
     }
 
     @Transactional
+    public void resolveOutboundCrash() {
+        List<Timestamp> crashed = operationDao.getAndResolveCrashed(OperationDao.EfgsOperationDirection.OUTBOUND);
+
+        if (!crashed.isEmpty()) {
+            String sql = "update en.diagnosis_key set sent_to_efgs = null, retry_count = 0 where sent_to_efgs in (:timestamp)";
+            jdbcTemplate.update(sql, Map.of(
```
jyrkija: These extra linebreaks look funny to my eye when this wouldn't even be a long row as a single-liner.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
```diff
@@ -1,161 +1,127 @@
...
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static java.util.Objects.requireNonNull;
 
 @Repository
 public class OperationDao {
+    public enum EfgsOperationState {STARTED, FINISHED, ERROR}
```
jyrkija, line 23: This enums are used in a couple of places and their names clearly communicate enough context on their own (without the wrapping OperationDao...) -> I'd move them outside this class.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
```diff
@@ -1,161 +1,127 @@
...
-                "new_state", EfgsOperationState.FINISHED.name(),
-                "id", operationId,
-                "keys_count_total", keysCountTotal
-        )) == 1;
+        Map<String, Object> params = new HashMap<>();
+        params.put("new_state", EfgsOperationState.FINISHED.name());
+        params.put("id", operationId);
+        params.put("batch_tag", batchTag.orElse(null));
+        params.put("keys_count_total", keysCountTotal);
+        params.put("updated_at", new Timestamp(Instant.now().toEpochMilli()));
```
jyrkija, line 44: This could also be generated in the db

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
```diff
@@ -1,161 +1,127 @@
...
-    public void setStalledToError() {
-        String sql = "update en.efgs_operation set state = cast(:error_state as en.state_t), updated_at = :updated_at " +
-                "where state = cast(:started_state as en.state_t) and (now() > (updated_at + interval '10 minute')) ";
-        jdbcTemplate.update(sql, Map.of(
-                "error_state", EfgsOperationState.ERROR.name(),
-                "started_state", EfgsOperationState.STARTED.name(),
-                "updated_at", new Timestamp(Instant.now().toEpochMilli())
-        ));
+    public void markErrorOperation(long operationId, Optional<String> batchTag, Timestamp timestamp) {
+        String sql = "update en.efgs_operation set state = cast(:error_state as en.state_t), updated_at = :updated_at, " +
```
jyrkija: For clarity, you could break these sql-lines so each assignment (x = :x) is on a separate line when there's more than 1

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
```diff
@@ -1,161 +1,127 @@
...
 
-    public List<Long> getOutboundOperationsInError() {
-        String sql = "select id from en.efgs_operation " +
-                "where state = cast(:state as en.state_t) " +
-                "and direction = cast(:direction as en.direction_t) " +
-                "and run_count < :run_count";
+    public List<Timestamp> getAndResolveCrashed(EfgsOperationDirection direction) {
+        String sql = "update en.efgs_operation set state = cast(:new_state as en.state_t) " +
+                "where state = cast(:current_state as en.state_t) and direction = cast(:direction as en.direction_t) " +
+                "and updated_at < :max_age returning updated_at";
```
jyrkija: max_age doesn't communicate very clearly that it's a timestamp (not age). Maybe stalled_operation_limit or somesuch?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
jyrkija: Altthough, as I read further, it's clear that it's compared with Java time later on. So no point moving this in to DB time, unless all of that is moved as well.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceWithDaoTestIT.java
```diff
@@ -94,9 +99,69 @@ public void downloadKeys() {
...
+                .andExpect(method(HttpMethod.GET))
+                .andRespond(withStatus(HttpStatus.OK)
+                        .contentType(PROTOBUF_MEDIATYPE)
+                        .headers(getDownloadResponseHeaders())
+                        .body(serialize(transform(keys)))
+                );
+        LocalDate date = LocalDate.now(ZoneOffset.UTC);
+        try {
+            federationGatewayService.startInbound(date, Optional.of(TEST_TAG_NAME));
+        } catch (Exception e) {
```
jyrkija: For stronger assertion and clearer intent, just catch the one exception you're expecting

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceWithDaoTestIT.java
```diff
@@ -94,9 +99,69 @@ public void downloadKeys() {
...
+                        .body("")
+                );
+        LocalDate date = LocalDate.now(ZoneOffset.UTC);
+        try {
+            federationGatewayService.startInbound(date, Optional.of(TEST_TAG_NAME));
+        } catch (Exception e) {
+            assertEquals(1, operationDao.getInboundErrorBatchTags(date).get(TEST_TAG_NAME));
+        }
+        IntStream.rangeClosed(2, MAX_RETRY_COUNT).forEach(i -> {
+                    try {
```
jyrkija, line 158: Odd indentation

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, this is not very nice. I'll put that on todo list.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, that was on my mind also. I changed it to repeatable isolation level for now, which should do the trick if both nodes tries to do this concurrently, because we don't mind that other one just fails when the changed data has been committed. Anyway, let's discuss about this more if needed.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, I thought that also. I changed it into batch, it would be better. Good point.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: yep, it was a little bit. I changed that to oneliner.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: Fixed imports. I'll put that moving on the todo list. They're kind of part of operation, I'm not much of fan of globals. Let's see where this goes.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: Yep, this ia a little bit nasty part. It's still shaping, I think, let's see, but for now I let it be.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: Yep, I have done it now for that specific line. Check others later.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: yep, I changed that

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceWithDaoTestIT.java
solita-alperttiti: Yep, that's a good point. I'm planning to sanitize this test class later, but changed these in this point.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: And, put second thought for this. Repeatable read is probably too restrictive in this. I added for update skip locks and order by for select. Let's see how this ends.

PR #41, opened 2020-11-19
refined diagnosis key api to be more sensible with padded data

solita-alperttiti:
Some tuning to make rest-api more sensible.

### DIFF
--- a/exposure-notification/README.md
+++ b/exposure-notification/README.md
@@ -262,6 +262,7 @@ Status API is a one-call replacement for diagnosis key current & list fetches as
 * **Request Body:**
   * keys: List of TemporaryExposureKey objects, as per [the API docs](https://developers.google.com/android/exposure-notifications/verification-system#metadata)
     * <u>Note, that the key data bytes need to be base64 encoded</u>
+    * <u>Note, for visitedCountries and consentToShareWithEfgs value is boolean, use 1 for true and 0 for false</u>
   * Sample Body 
     ```json
     {
@@ -275,8 +276,8 @@ Status API is a one-call replacement for diagnosis key current & list fetches as
           "transmissionRiskLevel": 5,
           "rollingStartIntervalNumber":  123456,
           "rollingPeriod": 144,
-          "visitedCountries": "DE",
-          "consentToShareWithEfgs": true
+          "visitedCountries": {"DE":1,"IT":0},
+          "consentToShareWithEfgs": 1
         }] 
     }
     ```
--- a/exposure-notification/README.md
+++ b/exposure-notification/README.md
@@ -275,10 +275,10 @@ Status API is a one-call replacement for diagnosis key current & list fetches as
           "keyData": "/MwsNfC4Rgnl8SxV3YWrqA==",
           "transmissionRiskLevel": 5,
           "rollingStartIntervalNumber":  123456,
-          "rollingPeriod": 144,
-          "visitedCountries": {"DE":1,"IT":0},
-          "consentToShareWithEfgs": 1
-        }] 
+          "rollingPeriod": 144
+        }],
+      "visitedCountries": {"DE":1,"IT":0},
+      "consentToShareWithEfgs": 1
     }
     ```
 * **Success Response:**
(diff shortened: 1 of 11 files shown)


exposure-notification/README.md
```diff
@@ -275,8 +276,8 @@ Status API is a one-call replacement for diagnosis key current & list fetches as
           "transmissionRiskLevel": 5,
           "rollingStartIntervalNumber":  123456,
           "rollingPeriod": 144,
-          "visitedCountries": "DE",
```
jyrkija, line 278: Why are these here for every key? Shouldn't they be once for the whole body?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyService.java
```diff
@@ -66,7 +66,7 @@ private long getExportedKeyCount(List<TemporaryExposureKey> keys) {
                 getRiskBucket(symptomsOnset, utcDateOf10MinInterval(requestKey.rollingStartIntervalNumber)),
                 requestKey.rollingStartIntervalNumber,
                 requestKey.rollingPeriod,
-                requestKey.visitedCountries,
+                requestKey.visitedCountries.keySet(),
```
jyrkija: I expected to see the filtering of value != 1 here. Where is it?

jyrkija: Oh, right below. The field inside the object might as well be the set directly; if you're filtering in the parsing part (which is fine), the map structure is pointless after that.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/Validation.java
```diff
@@ -80,6 +81,14 @@ public static int validateRollingPeriod(int rollingPeriod) {
         return codes.stream().filter(code -> !code.equals("FI")).filter(Validation::validateISOCountryCode).collect(Collectors.toSet());
     }
 
+    public static Map<String, Boolean> validateISOCountryCodesWithoutFI(Map<String, Boolean> codes) {
+        return validateISOCountryCodesWithoutFI(codes.entrySet().stream().filter(Map.Entry::getValue)
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toSet()))
```
jyrkija: One level of processing ends here with the ending ) but the indentation remains the same. This could be made clearer with some well-placed linebreaks and indentation.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/v1/TemporaryExposureKeyRequest.java
```diff
@@ -15,31 +17,44 @@
...
+     * Increments of 10 minutes describing how long a key is valid
+     **/
     public final int rollingPeriod;
-    /** Set of visited countries in ISO-3166 alpha-2 format **/
-    public final Set<String> visitedCountries;
-    /** Consent to share data with efgs **/
+    /**
+     * Set of visited countries in ISO-3166 alpha-2 format
+     **/
+    public final Map<String, Boolean> visitedCountries;
```
jyrkija: Move to request-level, not key-level, as they're given just once.

jyrkija (changes requested):

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyService.java
solita-alperttiti: It's complicates e.g. tests, so that's why this is it in this way.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/v1/TemporaryExposureKeyRequest.java
solita-alperttiti: Ah, you are correct. It was already there, so didn't double check think that. Good point. I'll fix this.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/Validation.java
solita-alperttiti: For indents I'm using ideas defaults, it might be a good idea to make some project specific settings in some point, but not know. I'll let what I can do for this.

exposure-notification/README.md
solita-alperttiti: Yes, it should.

jyrkija (approved):

PR #42, opened 2020-11-19
Implementation for more robust fetch

solita-alperttiti:
Implementation for more robust fetch, until we get callback available and as another option for download

- tries to prevent unnecessary rest-calls
- scheduling support to fetch yesterday's keys after utc midnight
- some refactoring + tuning

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
@@ -9,6 +9,7 @@ import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static net.logstash.logback.argument.StructuredArguments.keyValue;
 
@@ -16,13 +17,13 @@ import static net.logstash.logback.argument.StructuredArguments.keyValue;
 public class FederationGatewaySyncProcessor {
 
     private static final Logger LOG = LoggerFactory.getLogger(FederationGatewaySyncProcessor.class);
-    private LocalDate lastInboundSyncFromEfgs;
+    private AtomicReference<LocalDate> lastInboundSyncFromEfgs;
 
     private final FederationGatewaySyncService federationGatewaySyncService;
 
     public FederationGatewaySyncProcessor(FederationGatewaySyncService federationGatewaySyncService) {
         this.federationGatewaySyncService = federationGatewaySyncService;
-        this.lastInboundSyncFromEfgs = LocalDate.now(ZoneOffset.UTC);
+        this.lastInboundSyncFromEfgs = new AtomicReference<>(LocalDate.now(ZoneOffset.UTC));
     }
 
     @Scheduled(initialDelayString = "${covid19.federation-gateway.upload-interval}",
@@ -37,11 +38,11 @@ public class FederationGatewaySyncProcessor {
             fixedRateString = "${covid19.federation-gateway.download-interval}")
     private void runImportFromEfgs() {
         LocalDate today = LocalDate.now(ZoneOffset.UTC);
-
-        if (today.isAfter(lastInboundSyncFromEfgs)) {
+        LocalDate last = lastInboundSyncFromEfgs.get();
+        if (today.isAfter(last)) {
             LOG.info("Starting scheduled import from efgs.");
-            federationGatewaySyncService.startInbound(lastInboundSyncFromEfgs, Optional.empty());
-            lastInboundSyncFromEfgs = today;
+            federationGatewaySyncService.startInbound(last, Optional.empty());
+            lastInboundSyncFromEfgs.set(today);
             LOG.info("Scheduled import from efgs finished.");
         }
     }
@@ -52,7 +53,7 @@ public class FederationGatewaySyncProcessor {
         LOG.info("Starting scheduled efgs error handling.");
         federationGatewaySyncService.resolveCrash();
         federationGatewaySyncService.startOutbound(true);
-        federationGatewaySyncService.startInboundRetry(lastInboundSyncFromEfgs);
+        federationGatewaySyncService.startInboundRetry(lastInboundSyncFromEfgs.get());
         LOG.info("Scheduled efgs error handling finished.");
     }
 }
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
@@ -5,7 +5,6 @@ import org.slf4j.LoggerFactory;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 
-import java.time.Duration;
 import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.util.Optional;
@@ -17,36 +16,43 @@ import static net.logstash.logback.argument.StructuredArguments.keyValue;
 public class FederationGatewaySyncProcessor {
 
     private static final Logger LOG = LoggerFactory.getLogger(FederationGatewaySyncProcessor.class);
+    private LocalDate lastInboundSyncFromEfgs;
 
-    private final FederationGatewayService federationGatewayService;
+    private final FederationGatewaySyncService federationGatewaySyncService;
 
-    public FederationGatewaySyncProcessor(FederationGatewayService federationGatewayService) {
-        this.federationGatewayService = federationGatewayService;
+    public FederationGatewaySyncProcessor(FederationGatewaySyncService federationGatewaySyncService) {
+        this.federationGatewaySyncService = federationGatewaySyncService;
+        this.lastInboundSyncFromEfgs = LocalDate.now(ZoneOffset.UTC);
     }
 
     @Scheduled(initialDelayString = "${covid19.federation-gateway.upload-interval}",
             fixedRateString = "${covid19.federation-gateway.upload-interval}")
     private void runExportToEfgs() {
         LOG.info("Starting scheduled export to efgs.");
-        Set<Long> operationIds = federationGatewayService.startOutbound(false);
+        Set<Long> operationIds = federationGatewaySyncService.startOutbound(false);
         LOG.info("Scheduled export to efgs finished. {}", keyValue("operationId", operationIds));
     }
 
     @Scheduled(initialDelayString = "${covid19.federation-gateway.download-interval}",
             fixedRateString = "${covid19.federation-gateway.download-interval}")
     private void runImportFromEfgs() {
-        LOG.info("Starting scheduled import from efgs.");
-        federationGatewayService.startInbound(LocalDate.now(ZoneOffset.UTC), Optional.empty());
-        LOG.info("Scheduled import from efgs finished.");
+        LocalDate today = LocalDate.now(ZoneOffset.UTC);
+
+        if (today.isAfter(lastInboundSyncFromEfgs)) {
+            LOG.info("Starting scheduled import from efgs.");
+            federationGatewaySyncService.startInbound(lastInboundSyncFromEfgs, Optional.empty());
+            lastInboundSyncFromEfgs = today;
+            LOG.info("Scheduled import from efgs finished.");
+        }
     }
 
     @Scheduled(initialDelayString = "${covid19.federation-gateway.error-handling-interval}",
             fixedRateString = "${covid19.federation-gateway.error-handling-interval}")
     private void runErrorHandling() {
         LOG.info("Starting scheduled efgs error handling.");
-        federationGatewayService.resolveCrash();
-        federationGatewayService.startOutbound(true);
-        federationGatewayService.startInboundRetry(LocalDate.now(ZoneOffset.UTC).minusDays(1));
+        federationGatewaySyncService.resolveCrash();
+        federationGatewaySyncService.startOutbound(true);
+        federationGatewaySyncService.startInboundRetry(lastInboundSyncFromEfgs);
         LOG.info("Scheduled efgs error handling finished.");
     }
 }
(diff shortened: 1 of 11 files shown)


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
```diff
@@ -17,36 +16,43 @@
 public class FederationGatewaySyncProcessor {
 
     private static final Logger LOG = LoggerFactory.getLogger(FederationGatewaySyncProcessor.class);
+    private LocalDate lastInboundSyncFromEfgs;
```
jyrkija: To maintain the class as thread-safe, I'd throw this into a private final AtomicReference<LocalDate>. Of course in practice, it likely won't ever matter here, but in theory, the next pass might not be done by the same thread.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncService.java
```diff
@@ -76,27 +76,31 @@ private void doInbound(String date, Optional<String> batchTag) {
...
-
-            } else {
-                return Optional.empty();
-            }
-        } finally {
-            if (!finished) {
-                operationDao.markErrorOperation(operationId, localBatchTag);
+        Optional<Long> operationId = operationDao.startOperation(INBOUND, localBatchTag);
+
+        if (operationId.isPresent()) {
```
jyrkija: The if-else could be formatted as
`
return operationId.flatMap(id -> {
  ...
});
`

Or alternatively, even `return operationId.flatMap(this::tryDownload)` where signature is `Optional<DownloadData> tryDownload(long operationId)`

Both options avoid needing to .get() from the optional and the extra -else branch.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
```diff
@@ -60,6 +66,11 @@ public boolean finishOperation(FederationOutboundOperation operation, int keysCo
                 "keys_count_500", keysCount500,
                 "updated_at", new Timestamp(Instant.now().toEpochMilli())
         )) == 1;
+        LOG.info("Efgs sync finished. {} {} {}",
```
jyrkija: Maybe log the success-field too?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
```diff
@@ -93,8 +107,10 @@ public void markErrorOperation(long operationId, Optional<String> batchTag, Time
         ), Timestamp.class);
     }
 
-    public long startOperation(EfgsOperationDirection direction) {
-        return startOperation(direction, new Timestamp(Instant.now().toEpochMilli()));
+    public Optional<Long> startOperation(EfgsOperationDirection direction, Optional<String> batchTag) {
```
jyrkija, line 118: Didn't go through the use-places, but is there a need to combine these two fetches under one method? This looks like a typically multi-use function. Having a startOperation(direction) and startOperation(direction,non-optional tag) is typically cleaner and it would seem that when you're calling the method, you either have the tag or you don't -> you know what to call. 

Or is it intentional that when you have a tag and fail to find anything, you get the same result as when you don't have a tag at all? Maybe that orElseGet could be done in caller-side, if there aren't transaction things in the way.

jyrkija: Should this be @Transactional, btw?

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceSyncWithDaoTestIT.java
```diff
@@ -170,7 +182,7 @@ public void downloadRetryMaxLimit() {
 
     @Test
     public void downloadKeysTransmissionRiskLevel() {
-        List<TemporaryExposureKey> keys = FederationGatewayBatchUtil.transform(FederationGatewayBatchUtil.transform(
+        List<TemporaryExposureKey> keys = transform(transform(
```
jyrkija, line 185: transform transform. For the autobots :D Seriously though, maybe those function names could describe what/how they transform sot this would make more sense.

jyrkija: I don't know if it's the best convention but I've often settled to names like toX or X.from in transformations. The parameter-name, field name or function class then give the rest of the meaning.
For example (preferring the first one):
inputs = toInternalForm(externalFormat)
inputs = externalFormat.toMyInternalClass()
inputs = MyInternalClass.from(externalFormat)

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
solita-alperttiti: Yep, that is true, but even if it's run more than once as parallel, the value is only updated more than once. However, there might be some visibility issues for updates, so good point.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncService.java
solita-alperttiti: Yep, I know, for my eye it's just more mess...

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: added

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: There is only one statement on both calls, so it shouldn't be necessary? At least I think, it's not. However, there is no reason why it cannot be in there.

About api, it might be possible separate inbound and outbound, because the context is little bit different. I'll check this out later. about this tag-thingy it's fail safe for not to download everything from eu-server. Batches is an linked list and if you don't know the batch tag, then you just need to download the first batch and then you get the tag and also next tag, so I think it's cleaner in here.

exposure-notification/src/test/java/fi/thl/covid19/exposurenotification/efgs/FederationServiceSyncWithDaoTestIT.java
solita-alperttiti: This just to for tests, that I don't need to manually handle autogenerated fields from transform. It's visible from method definition what it will take as parameter and what you can except. Anyway, in actual code it is never used like that.

solita-alperttiti: Anyway, this is something which could be discussed more. I see your point. :)

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncService.java
solita-alperttiti: But I'll check this in some point.

solita-alperttiti: Tuned this a little bit. Maybe it's a little bit nicer now. It's good that someone makes noise of these. I must admit that it looked ugly also for me, but not too ugly in current situation. :)

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
solita-alperttiti: I decided to add @Transactional for all public methods, maybe it's easy to forgot it later if you add something. After second I also added it to inbound keys batch update, so it's all or nothing.

solita-alperttiti: It could also be at class level, but maybe it's good to be explicit in this case.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
solita-alperttiti: However, I checked that we have only @EnableScheduling defined in our configuration, so it seems that our scheduler is single-threaded. So, not much problem of concurrency then. :) Anyway, maybe we should define some thread pool because we have increasing count of jobs. Lets talk about this tomorrow.

_By default, will be searching for an associated scheduler definition: either a unique TaskScheduler bean in the context, or a TaskScheduler bean named "taskScheduler" otherwise; the same lookup will also be performed for a ScheduledExecutorService bean. If neither of the two is resolvable, a local single-threaded default scheduler will be created and used within the registrar._

solita-alperttiti: Although, our services seems not to be very timing related, so single thread might actually be enough. Just thinking if there is a risk that it get stuck. However, running more threads is a bad solution for that anyway. Anyway, we could add something like this in future in application.yml if there is need for that:
`spring.task.scheduling.pool.size=5`

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
```diff
@@ -5,48 +5,55 @@
...
     }
 
     @Scheduled(initialDelayString = "${covid19.federation-gateway.download-interval}",
             fixedRateString = "${covid19.federation-gateway.download-interval}")
     private void runImportFromEfgs() {
-        LOG.info("Starting scheduled import from efgs.");
-        federationGatewayService.startInbound(LocalDate.now(ZoneOffset.UTC), Optional.empty());
-        LOG.info("Scheduled import from efgs finished.");
+        LocalDate today = LocalDate.now(ZoneOffset.UTC);
+        LocalDate last = lastInboundSyncFromEfgs.get();
```
jyrkija, line 41: This could be done atomically with updateAndGet.
Also, now that I read it again, I started wondering about storing the integration state in memory like this. Shouldn't it be in the DB? Or how is this meant to behave when the service is restarted?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/OperationDao.java
jyrkija: Mostly I was thinking that does the first operation change something in the DB that should be rolled back if we get a disconnect in the middle of the second, etc. I don't think so, since the second is a fallback for when the first does nothing, so there should be nothing to roll back.

I wouldn't favor putting transactional to everything. I think the annotation is a good indication that multiple things happen here and the reader needs to pay attention. Most of the tasks here are so simple that there's no need for transactional, and that's always the first goal. Of course, the transactional doesn't actually hurt anything either, so it's fair to disagree.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
jyrkija: Yeah, a single scheduler should be fine for us. Only reason for adding more would be if for example the efgs part starts taking so long that we don't want other tasks (like cleaning keys from db) to wait for it. I don't think there's much risk of that. None the less, in services like this where everything is 99% stateless anyhow, I like to assume that I don't know the threading/call orders -> avoid state first and thread-safe the small bits that you might find needed.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncService.java
jyrkija: It does look cleaner to my eye :) The function now more clearly "does one thing" and the caller decides if it has the sufficient arguments to make the call in the first place.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/FederationGatewaySyncProcessor.java
solita-alperttiti: It doesn't really matter in this case. This is just meant to prevent too much downloading and we cannot prevent that entirely if we don't want to go too far with this. Anyway, actual fail safe is on the db-layer. So, it's ok to initialize it again on startup. And all nodes also. We're heading towards callback and when it is going to be activated this logic will be there just for special cases.

Also, in this case it's totally acceptable to let the last update to be written to the atomicreference, even if there is parallel processing involved, because that's what it's trying to representate, when the last operation ran.

About updateAndGet, not so sure about that, at least in case if you mean that we would pass the entire operation for it. There is some chance that updateAndGet would be called twice and that's something what we're trying to prevent in here. However, it might possible to use it at start and then if the actual operation is failing, switch it back to old value. Maybe that was what you meant. However, I'll let it be for now, but we could tune this later.

solita-alperttiti: Anyway, I changed this a little bit, so if we choose to use multiple threads in some point it will be cleaner. This is a little it over-engineering if we count the facts mentioned, but it's better. You are correct.

solita-alperttiti: Actually, after a good night sleep, I switched back to square one with two new powers, volatile for instance variable and running job with delay. Code is nice and clear now. :)

PR #43, opened 2020-11-20
Callback initialization based on properties

solita-alperttiti:


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/Callback.java
```diff
@@ -0,0 +1,12 @@
+package fi.thl.covid19.exposurenotification.efgs;
+
+public class Callback {
+
+    public final String callbackId;
+    public final String url;
+
+    public Callback(String callbackId, String url) {
+        this.callbackId = callbackId;
```
jyrkija: requireNonNull

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/CallbackInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+
+import static java.util.Objects.requireNonNull;
+import static net.logstash.logback.argument.StructuredArguments.keyValue;
+
+@Component
+@Profile("!dev & !test")
```
jyrkija, line 19: While this is fine, it's generally better that profiles configure features than that features know what profiles they associate with. Here, it would mean binding this to some callback-enabled property (there was going to be one anyhow, right?) and set that to false in dev- and test- application.yml:s. That way, the dependency direction is flipped and the pretty implementation class doesn't have to know about possible runtime-environments.

solita-alperttiti: Yep, but in this case we then would need two enabled switches, one which enables whole bean and one which enables callback. It seemed kind of messy. Anyway, this is matter of opinion, I might chańge it in future versions.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/efgs/Callback.java
solita-alperttiti: I added this. I'm starting to think that it's better having traditional nullpointer than this kind of boilerplate.... :)

PR #50, opened 2021-01-11
Support exposure windows aka V2

solita-alperttiti:
This pull request resolves # COVID-495, COVID-504, COVID-101

This is not probably in final state, but can be reviewed anyway.


exposure-notification/README.md
```diff
@@ -219,6 +219,70 @@ If a config is returned, it should be stored and used instead.
...
+              "reportTypeWeightRecursive":0.0,
+              "infectiousnessWeightStandard":1.0,
+              "infectiousnessWeightHigh":1.5,
+              "attenuationBucketThresholdDb":[55,70,80],
+              "attenuationBucketWeights":[2.0,1.0,0.25,0.0],
+              "daysSinceExposureThreshold":10,
+              "minimumWindowScore":1.0,
+              "minimumDailyScore":900,
+              "daysSinceOnsetToInfectiousness": {
+                  "-1":"HIGH",
```
jyrkija, line 263: Odd ordering of values. Hashmap? Treemap would likely give a more readable representation.

exposure-notification/src/main/resources/db/migration/V04.02__diagnosis_key_v2.sql
```diff
@@ -0,0 +1 @@
+alter table en.diagnosis_key add submission_interval_v2 int not null;
```
jyrkija: Doesn't this break without a default value (on non-empty db)?

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileService.java
```diff
@@ -78,9 +92,24 @@ public int cacheMissingBatchesBetween(int fromInterval, int untilInterval) {
...
+        if (intervals.current == intervals.last) {
+            batches = Stream.concat(batches, getDemoBatchIdV2(intervals.current).stream());
+        }
+        return batches.filter(previous::isBefore).collect(Collectors.toList());
+    }
+
     public Optional<BatchId> getDemoBatchId(int currentInterval) {
         int count = dao.getKeyCount(currentInterval);
-        return count > 0 ? Optional.of(new BatchId(currentInterval, Optional.of(count))) : Optional.empty();
+        return count > 0 ? Optional.of(new BatchId(currentInterval, Optional.of(Integer.parseInt("" + currentInterval + count)))) : Optional.empty();
```
jyrkija, line 107: Seems a little odd to concatenate the numbers without a separator. Guess it works well enough for the demo feature, though. This is probably to avoid having to do firewall configuration changes for a different ID format?

jyrkija (approved):

exposure-notification/README.md
solita-alperttiti: It's json, so order is not significant by specification. :) Anyway, map is coming straight from the resultset, so there's not much to do with types. However, some sorting could be added to transform operation. I'll check that in some point.

exposure-notification/src/main/resources/db/migration/V04.02__diagnosis_key_v2.sql
solita-alperttiti: good point. Fixed.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileService.java
solita-alperttiti: Yep, it's more like just to get it done in some manner. Reasoning behind this is more like that we don't entirely separate logic for demo batches. Anyway, good point and I'll check this in some point if it could be improved.

exposure-notification/README.md
solita-alperttiti: fixed

PR #59, opened 2021-04-13
Dummy keys generation for small batches

solita-alperttiti:
This pull request resolves COVID-506 .

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
@@ -20,7 +20,7 @@ import java.util.*;
 import java.util.function.IntFunction;
 import java.util.stream.Collectors;
 
-import static fi.thl.covid19.exposurenotification.diagnosiskey.IntervalNumber.from24hourToV2Interval;
+import static fi.thl.covid19.exposurenotification.diagnosiskey.IntervalNumber.toV2Interval;
 import static fi.thl.covid19.exposurenotification.efgs.util.DummyKeyGeneratorUtil.*;
 import static fi.thl.covid19.exposurenotification.efgs.util.CommonConst.MAX_RETRY_COUNT;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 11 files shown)


solita-alperttiti:
With this initial implementation we got different set of dummy keys for national and efgs batches. So, if this is going to be the case in final implementation it might be possible to identify dummy keys by comparing batches from different national backends, or is it? So, we'll probably need to store dummy keys on our side in some manner. However, national and efgs batch generation are separate processes, so we're probably going to need some sort of synchronization in here. Anyway, let's discuss about our options in some point.

solita-alperttiti:
Draft version 2. After a second thought, let's generate dummies on every efgs outbound operation when necessary and insert into the db where dummies will be eventually fetched into the national batches.

solita-alperttiti:
Draft version 3. Generate dummies also, if necessary, for national batches. E.g. id there is only keys with no consent to efgs. However, concurrency still needs some checks.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
```diff
@@ -109,39 +112,53 @@ public int getKeyCountV2(int intervalV2) {
...
         // We should not have invalid data in the DB, but if we do, pass by it and move on
-        return jdbcTemplate.query(sql, params, (rs, i) -> mapValidKey(interval, rs, i))
+        List<TemporaryExposureKey> keys = jdbcTemplate.query(sql, params, (rs, i) -> mapValidKey(interval, rs, i))
                 .stream().flatMap(Optional::stream).collect(Collectors.toList());
+
+        if (keys.isEmpty() || keys.size() >= BATCH_MIN_SIZE) {
+            return keys;
+        } else {
+            List<TemporaryExposureKey> dummyKeys = generateDummyKeys(BATCH_MIN_SIZE - keys.size(), false, from24hourToV2Interval(interval), Instant.now());
+            batchInsert(dummyKeys, Optional.empty());
```
jyrkija: I dislike that a getSomething method has a side effect of inserting stuff into the database. Perhaps this could be a separate function, for example padKeysForInterval(int interval) that just does the count-check and inserts dummies. That can then be called before the actual fetch for the same effect but keeping the concerns separate (and separately testable).

jyrkija: That would also avoid having the near-identical code in all of these fetches

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/TemporaryExposureKey.java
```diff
@@ -78,6 +78,10 @@ public TemporaryExposureKey(String keyData,
         this.submissionIntervalV2 = submissionIntervalV2;
     }
 
+    public String getKeyData() {
```
jyrkija, line 81: It's a public field - why the getter?

jyrkija: Oh right, function reference

jyrkija (approved):
Looks like it does what it says on the tin. As for synchronization, I was thinking maybe the process of dummy-generation should be a separate step in the maintenance routine.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/diagnosiskey/DiagnosisKeyDao.java
solita-alperttiti: Yep, good points and good discussions on other channels. I pushed kind of compromise solution just a moment ago.

jyrkija (approved):

PR #62, opened 2021-04-15
Dependencies upgraded

solita-alperttiti:
- spring boot to version 2.4.5
- removed flyway-core version and use spring defined version instead
- patch updates
- dependency check upgraded


jyrkija (commented):
:+1:

PR #63, opened 2021-04-21
Fixed issue: batch id might overflow with large batches and demo mode enabled

solita-alperttiti:

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileService.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileService.java
@@ -104,12 +104,14 @@ public class BatchFileService {
 
     public Optional<BatchId> getDemoBatchId(int currentInterval) {
         int count = dao.getKeyCount(currentInterval);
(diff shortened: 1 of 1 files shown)


exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileService.java
```diff
@@ -164,4 +170,14 @@ public BatchId getLatestBatchIdV2(BatchIntervals intervals) {
             return BatchFileFactory.createBatchFile(signatureConfig, signingKey, metadata, keys);
         }
     }
+
+    private int generateDemoId(int keyCount, int base) {
+        if (keyCount < base) {
+            return keyCount;
+        } else if (keyCount % base == 0) {
+            return base - 1;
```
jyrkija, line 178: So it gives the 9999 id twice? Why? I assume the point is to avoid overflowing the integer with an extra digit, but I don't see how repeating the last number is better than starting again from 0.

solita-alperttiti: it uses it every time modulo is zero, which is not nice as counter. Most likely in practical cases it will be not used twice (or three-times, etc), because it's not so common that key count increases only with one key. In practice this affects only to our test environments where total keys count is millions.

PR #64, opened 2021-04-30
WIP: Allow reading the same file from multiple threads in the same JVM (maintaining file locks)

jyrkija:
The file locks in batchfilestorage are needed so that no-one ends up reading files that someone is still in the middle of writing. However, the read lock cannot be obtained multiple times and in some cases the batch reads can come through caches in a burst, ending up reading the file multiple times. This isn't a big problem, as some will succeed, others fail and the failed attempts will be retried, this time finding the file in the cache. 

However, it's cleaner to allow reading the file in parrallel. This PR tries to accomplish that by maintaining a single filechannel (thread safe in itself) with a single lock through multiple reads. It only gets closed once all reads are done.


jyrkija:
This is an over-complicated solution to an issue that is better resolved by https://github.com/THLfi/koronavilkku-backend/pull/65
Closing as redundant.

PR #65, opened 2021-04-30
Synchronize multithreaded file reads by using a cache -> IO once and avoid multi-locking

jyrkija:
Solves the same issue as https://github.com/THLfi/koronavilkku-backend/pull/64 but cleaner

### DIFF
--- a/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileStorage.java
+++ b/exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileStorage.java
@@ -3,6 +3,7 @@ package fi.thl.covid19.exposurenotification.batch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.stereotype.Service;
 
 import java.io.File;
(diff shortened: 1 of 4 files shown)


exposure-notification/src/main/resources/application.yml
```diff
@@ -75,6 +75,7 @@ covid19:
     data-cache:
       enabled: true
       status-duration: PT5M
+      file-duration: PT10S
   maintenance:
```
solita-alperttiti, line 79: I'm thinking should we cache longer? Batch with the same batchId shouldn't change and we'll need to be prepared that all available batches are in the memory at the same time anyway.

exposure-notification/src/main/java/fi/thl/covid19/exposurenotification/batch/BatchFileStorage.java
```diff
@@ -117,6 +120,8 @@ public void addBatchFile(BatchId batchId, byte[] data) {
         }
     }
 
+    // Note: This also caches empty value (rejecting it is not compatible with sync) so the time should be short.
+    @Cacheable(value = "batch-file", sync = true)
     public Optional<byte[]> readBatchFile(BatchId batchId) {
```
solita-alperttiti, line 125: This seems to be exactly what we are looking for.

solita-alperttiti (changes requested):
Looks good and straightforward. I'll mark it as request changes for further discussions. Even it can be merged in the current shape too.

exposure-notification/src/main/resources/application.yml
jyrkija: We certainly can increase this, but I was thinking that it could cause more trouble than help. Mostly we get this situation as a single burst and then it disappears as the load balancers caches the file. After that, this cache becomes irrelevant.

It's true, the batch won't change, but if it hasn't been written yet, it may get written very soon and the cache will continue returning empty until it expires, meaning that the other readers each continue building the batch directly from DB... which would also work and shouldn't even happen as the batch should be there already. So either way works, I just figured that it's not needed after the short interval and this means that the situation is re-checked from disk without too much delay.

solita-alperttiti: Yep, let's keep in in that way and we can tune it later if needed.

solita-alperttiti (approved):

PR #67, opened 2021-05-14
Response with code 502 when sms sending fails

solita-alperttiti:

### DIFF
--- a/publish-token/README.md
+++ b/publish-token/README.md
@@ -86,7 +86,7 @@ Deprecated components will be removed in next api-version.
     }
     ```
 * **Failure Response:**
-    * Error when communicating with sms-gateway
+    * Error when communicating with SMS-gateway
     * Status: 502 Bad Gateway
     
 ### Fetch Tokens Created by a User
--- a/publish-token/README.md
+++ b/publish-token/README.md
@@ -85,6 +85,9 @@ Deprecated components will be removed in next api-version.
       "validThroughTime": "2020-07-02T15:13:20.050472Z"
     }
     ```
+* **Failure Response:**
+    * Error when communicating with sms-gateway
+    * Status: 502 Bad Gateway
     
 ### Fetch Tokens Created by a User
 Request tokens created by a single user, so that the user may re-check the code in case of mistakes or misreads.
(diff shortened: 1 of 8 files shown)


publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationController.java
```diff
@@ -48,7 +48,12 @@ public PublishToken generateToken(@RequestHeader(name = SERVICE_NAME_HEADER) Str
                     requestService,
                     request.requestUser,
                     request.symptomsExist);
-            request.patientSmsNumber.ifPresent(number -> smsService.send(number, token));
+            request.patientSmsNumber.ifPresent(number -> {
+                        if (!smsService.send(number, token)) {
+                            throw new IllegalStateException("SMS send failed");
```
jyrkija: Certainly a clean implementation. However, this does mean that the user has no other option than to try again. If there's an issue in the SMS service, there's no way to get out a valid code. Also, since the code is already generated at this point (just not returned), the user might aggressively retry, leading to many valid codes in backend. Of course, they'll expire pretty soon, so it's likely not a big deal.

Overall, I guess this is fine if SMS-service errors are expected to be rare and short-lived. It would be nicer to return both the code and SMS status, but that's obviously a much bigger change and requires changes on the caller side as well.

jyrkija (approved):

publish-token/src/main/java/fi/thl/covid19/publishtoken/generation/v1/PublishTokenGenerationController.java
solita-alperttiti: Yep, however user can create code without sms-number even if there is issues with sms-service. Anyway, while with human users, it's probably not a big problem have some unnecessary tokens, with bot-clients it could be. So, I have added invalidation functionality for erroneus tokens.

So, in case there is error when sending sms user needs to:

a) try again and wait the problem to be fixer
b) try again without phone number and deliver token to end-user with other means

I have also added error counter for more robust monitoring in server side.

jyrkija (approved):

jyrkija (approved):

publish-token/README.md
```diff
@@ -85,6 +85,9 @@ Deprecated components will be removed in next api-version.
       "validThroughTime": "2020-07-02T15:13:20.050472Z"
     }
     ```
+* **Failure Response:**
+    * Error when communicating with sms-gateway
```
jyrkija: SMS

jyrkija (approved):
Nice & clean :+1:

PR #79, opened 2022-01-21
DEPENDENCY_UPGRADE:

solita-alperttiti:
* spring boot to version 2.5.9
    * change postgresql to use version from spring-boot
    * git-commit-id-plugin relocation (no functional changes based on release notes)
    * cache2k to version 2.4.1.Final
    * logstash-logback-encoder to version 7.0.1
    * introduced commons-lang3 as dependency and use it in imports instead of version provided by earlier logstash-logback-encoder
    * owasp-dependency-check to version 6.5.3


jyrkija (approved):
:+1:

PR #80, opened 2022-03-04
Localize EOL configuration values (as opposed to just labels) to better support textual ones

jyrkija:


solita-alperttiti (approved):
Looks good!

